PR #4, opened 2026-07-22
Add Fable/premium usage tracking and master account list

rjwalters:
Adds Fable-tier (premium/overage) usage tracking and a master account-list
file that auto-loads at launch.

## What's included
- **Fable/premium usage tracking** — a separate, slower-cadence probe
  (`probeFableDue()` in OAuthPoller) that reads the premium/overage rate-limit
  headers; new `fable` and `extra` columns in the summary table (popover
  widened to 818px to fit them), with `fablePercent`/`extraUsage` sort keys.
- **Master account list** — `~/.claude-monitor/accounts.env` (+ optional
  `accounts.local.env`) loaded additively at launch via
  `syncFromAccountFiles()` / `syncThenRefreshAll()`; documented in the README.
- Store/schema and API-client changes to persist and expose the new fields.

## Notes
- Branched before the recent `main` work; `main` was merged in and the result
  builds clean (`swift build`). The auto-fit popover height and `ranking.json`
  exporter from `main` are preserved and coexist with the new Fable columns.
- No `Closes #N` — this is a feature branch, not tied to a tracked issue.


rjwalters:
✅ **Approved** — `loom:pr`

Thorough review of the Fable/premium usage tracking + master account-list feature (~450 lines across 6 files). CI `build` check passes, merge state is CLEAN, and `swift build` passes locally.

**Correctness — probe cadence & header parsing**
- `probeFableDue()` gates each account on `fableProbeInterval` (1200s → ~3×/hr) via `lastFableProbeTimes[accountId]`, independent of the Haiku poll. Correct staggering logic (`last == nil || elapsed >= interval`).
- `rawProbe()` never throws — network/HTTP failures return `(0, [:])` and are skipped (`if status > 0`), so a missing/absent Fable response can't crash or write junk.
- Header parsing is fully defensive: `Double(headers[...] ?? "")` guards, `extractAnthropicHeaders` filters to the `anthropic-` prefix and lowercases keys, and unknown keys are logged once/process (`loggedUnknownKeys`) so an API change surfaces instead of passing silently.
- `ExtraUsageState` cleanly models the messy real-world states (unknown/off/empty/active/ready/percent); the `.percent` remaining is only shown when the budget is actually metered (`used > 0`), matching the documented "unlimited balance never meters" behavior. Sort keys (`fablePercent`, `extraUsageUrgency`) sort nil/no-probe last as intended.

**Security — live `sk-ant-oat01-*` tokens**
- Tokens are never logged: `rawProbe` uses the token only as the `Authorization: Bearer` request header; `syncFromAccountFiles`/`importFromEnvString` log account *counts* and emails only, never token values.
- `probe_snapshots` stores only `anthropic-*` *response* headers — request auth headers are never captured, so no token leaks into the DB beyond the existing credential store.
- README documents `chmod 600` for `accounts.env`/`accounts.local.env` and the local-overrides-master merge semantics. (Advisory only — the app reads but doesn't enforce file perms; acceptable, non-blocking.)

**Backward-compatible schema**
- New data lives in a separate `probe_snapshots` table (`CREATE TABLE/INDEX IF NOT EXISTS`), run unconditionally by `ensureDatabase()` on every launch — existing `usage.db` upgrades cleanly with no `ALTER`. The new `UsageRecord` fields default to `nil` and are populated from that table, so `usage_history` needs no column migration.

**Coexistence with recent `main` work**
- Only `popoverWidth` (690→818) changed to fit the two new columns; the auto-fit `PopoverHeightManager` height logic and `RankingExporter`/`ranking.json` are untouched and coexist correctly. Additive account loading (`syncThenRefreshAll`) never removes existing accounts.

Clean, well-documented, defensively-coded feature. Nice work.

PR #9, opened 2026-07-28
Add headless Linux mode: portable core + UI-less poll loop daemon

### LINKED ISSUE #8: Headless Linux mode: run the usage poller as a daemon on cloud loom worker hosts
## Context

Loom is piloting multi-host scaling (rjwalters/loom#3979): cloud Linux worker hosts running the loom-daemon, each carrying the **full** OAuth account pool rather than a pinned subset. On the Mac hosts, claude-monitor is the source of truth for token availability — it polls the usage endpoints into `~/.claude-monitor/usage.db`, and `loom-daemon tokens import-from-monitor` (loom #4106) reads that SQLite directly, while `~/.claude-monitor/accounts.env` seeds the pool via `loom-tokens bootstrap`.

On a Linux worker host today (first one provisioned 2026-07-28: Ubuntu 24.04 on EC2), that whole layer is absent — claude-monitor is a macOS menu-bar app. The worker falls back to `loom-daemon tokens check`'s own HTTP probe, which works but gives each host an independent, point-in-time view instead of the continuously-polled history the Mac side gets, and `import-from-monitor` has nothing to import.

## Request

A headless mode — no menu bar, no windows — that runs the existing poller loop as a plain daemon on Linux:

- `claude-monitor --headless` (or a separate small binary/entry point) that reads the same `~/.claude-monitor/accounts.env`, polls on the same cadence, and writes the same `usage.db` schema so `loom-daemon tokens import-from-monitor` works unchanged.
- Runnable under systemd (sample unit file or docs snippet).
- No GUI dependencies linked in that mode (the existing `dist/cli.js` CLI suggests some of this split already exists; note #5 reports it currently fails at startup — fixing that may be a natural first step).

- [ ] On a Linux host with only `accounts.env` present, the headless poller populates `usage.db` on the normal cadence.
- [ ] `loom-daemon tokens import-from-monitor` on that host imports it successfully.
- [ ] Sample systemd unit (or documented `systemd-run` invocation) in the README.

rjwalters:
## Summary

Makes the Swift package build and run on Linux hosts as a headless daemon, as required for the Loom system (#8). Same poll loop as the menubar app — account-file sync, 10-minute usage pings, 20-minute Fable probes — writing the same `~/.claude-monitor/usage.db` and `ranking.json`, with no UI.

- UI sources fenced with `#if os(macOS)`; the portable core (API client, poller, store, SQLite wrapper, exporter, logger) has no AppKit/SwiftUI/Combine/os.Logger dependency
- New `HeadlessRunner` poll loop; Linux entry via `HeadlessMain.swift`, macOS testing via `ClaudeMonitor --headless`; flags `--once`, `--interval <sec>`, `--version`; account list files are re-imported when they change (a daemon has no relaunch moment)
- `CSQLite` system-library target maps Linux libsqlite3; macOS keeps the SDK `SQLite3` module; `LinuxCompat.swift` shims `ObservableObject`/`@Published`
- Bug fixed en route: rate-limit headers are now parsed from the lowercased captured map — `allHeaderFields` subscript lookups are case-sensitive on Linux and would have silently dropped every utilization value
- Sample systemd user unit at `scripts/claude-monitor.service`; README + CLAUDE.md document the Linux build; CHANGELOG entry added

## Verification

- macOS debug and release builds pass; `--headless --once` polled all 8 configured accounts end-to-end and refreshed `ranking.json`
- Linux (`swift:6.1` Docker + `libsqlite3-dev`): clean build; fresh-home run creates the full DB schema and exits 0; real-token run completed import → authenticated ping (org/utilization parsed correctly) → Fable probe 200 → complete `ranking.json`

Closes #8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Headless mode / Linux support.** The package now builds on Linux
+  (`libsqlite3-dev` + Swift toolchain) as a UI-less daemon for Loom hosts: the
+  same poll loop (account-file sync, 10-min usage pings, 20-min Fable probes)
+  writing the same `~/.claude-monitor/usage.db` and `ranking.json`. On macOS
+  the loop is reachable via `ClaudeMonitor --headless`. Flags: `--once`,
+  `--interval <sec>`, `--version`; account list files are re-imported on
+  change while running. Includes a sample systemd unit
+  (`scripts/claude-monitor.service`). UI sources are fenced with
+  `#if os(macOS)`; a new `CSQLite` system-library target maps Linux's
+  libsqlite3, and rate-limit headers are now parsed from the lowercased
+  captured map (Linux `allHeaderFields` lookups are case-sensitive).
+
 - **App icon.** Custom gauge-dial icon (generated with
   [Imagine](https://github.com/rjwalters/imagine)); `Assets/AppIcon.icns` is
   now bundled by `build-macos-app.sh` and referenced via `CFBundleIconFile`.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,10 +13,21 @@
   open /Applications/ClaudeMonitor.app
   ```
 
+## Headless / Linux
+
+- The same package builds on Linux (`swift build` with `libsqlite3-dev` installed) as a headless daemon for Loom hosts — no UI, same poll loop, same `usage.db`/`ranking.json` outputs. See README "Headless Mode / Linux".
+- UI sources are fenced with `#if os(macOS)`; portable core must stay free of AppKit/SwiftUI/Combine/os.Logger. On Linux, `LinuxCompat.swift` shims `ObservableObject`/`@Published`, and `CSQLite/` maps the system libsqlite3 (macOS uses the SDK's `SQLite3` module).
+- Entry points: `main.swift` (macOS, dispatches to `HeadlessRunner` on `--headless`) and `HeadlessMain.swift` (Linux, always headless). Flags: `--once`, `--interval <sec>`, `--version`.
+- Parse response headers via `extractAnthropicHeaders` (lowercased map), never `allHeaderFields` subscripts — those are case-sensitive on Linux.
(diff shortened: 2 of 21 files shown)


rjwalters:
## Judge Review — APPROVED

**Verified:**
- **CI**: build check passes; merge state CLEAN.
- **macOS build**: clean `swift build` on the PR branch; `--headless --version` / `--help` work (v1.15.0).
- **Linux build** (swift:6.1 Docker + libsqlite3-dev): clean build; fresh-home `--once` run creates the full schema (accounts, oauth_credentials, probe_snapshots, settings, usage_history), writes `ranking.json` + `debug.log`, logs to stdout, and exits 0 — so `loom-daemon tokens import-from-monitor` sees the unchanged schema, per issue #8's acceptance criteria.
- **Platform fences**: every AppKit/SwiftUI/Charts import sits inside an `#if os(macOS)` file; `import os` is fully removed; `FoundationNetworking` conditionally imported where URLSession types are used.
- **LinuxCompat shim**: sound — grep confirms the portable core never uses `$projection` or `objectWillChange`, so the storage-only `@Published`/`ObservableObject` stand-ins suffice; macOS still gets real Combine.
- **Header-parsing fix**: `extractAnthropicHeaders` lowercases keys and all reads (ping, org-id validation, raw probes) now go through it — a genuine Linux correctness fix with no macOS behavior change (the same map was already stored in `rawHeaders`).
- **No macOS regressions**: `pollInterval` let→var keeps the 600s default; `exportNow()` is additive (`queue.sync`); `loadFromDatabase` publishes synchronously so `logSummary` reads fresh state; `@main` dispatch preserves the original app path.
- **Docs**: README/CLAUDE.md/CHANGELOG claims match the code (flags, mtime-based account-file re-import, CSQLite target, systemd unit path); CHANGELOG entry is correctly under `[Unreleased]`.

**Non-blocking nits** (fine to fix in a follow-up or ignore):
1. `LinuxCompat.swift` comment mentions `UpdateChecker` — no such file exists in Sources.
2. `scripts/claude-monitor.service` is a user unit ordered `After=network-online.target`, which doesn't exist in the systemd user manager — the ordering is a harmless no-op (Wants=, not Requires=), but a comment or `Restart=on-failure` reliance note could clarify.

Approving — relabeling `loom:review-requested` → `loom:pr`.

PR #17, opened 2026-07-30
fix: drop no-op network-online ordering from user systemd unit

### LINKED ISSUE #14: claude-monitor.service: network-online.target ordering is a no-op in the systemd user manager
## Context
Review nit from PR #9: `scripts/claude-monitor.service` is installed as a systemd *user* unit (`systemctl --user`), but `After=network-online.target` / `Wants=network-online.target` reference a target that only exists in the *system* manager — in the user manager the ordering is a harmless no-op. Harmless today (the poller retries transient network errors), but the unit implies a guarantee it doesn't provide.

- [ ] Either drop the `network-online.target` lines from the user unit, or replace the install docs with a system-unit variant where the ordering works
- [ ] A comment in the unit (or README note) explains the choice so it doesn't get "fixed" back

rjwalters:
## Summary
`scripts/claude-monitor.service` is installed as a `systemctl --user` unit, but `network-online.target` has no provider in the systemd user manager — the `After=`/`Wants=network-online.target` lines were silently ignored, implying an ordering guarantee the unit never actually provided.

## Changes
- Removed `After=network-online.target` and `Wants=network-online.target` from the `[Unit]` section
- Added an explanatory comment in their place noting this is a user unit (no network-online provider in the user manager) and that the poller's own retry logic plus `Restart=on-failure` already cover startup-before-network

| Drop `network-online.target` lines from the user unit (or ship a system-unit variant) | Done | Chose Option 1 per Curator/Champion recommendation; both `After=` and `Wants=network-online.target` lines removed |
| A comment explains the choice so it isn't "fixed" back | Done | New `[Unit]` comment explains the user-vs-system-manager distinction and why the poller doesn't need it |
| Unit file no longer references `network-online.target` anywhere | Done | `grep -n network-online scripts/claude-monitor.service` returns no matches |
| Install instructions still match the user-unit workflow | Done | Header comment (`systemctl --user …`) unchanged, still accurate |
| README | N/A | README has no mention of `network-online.target`; unit-file comment is the single source of truth per Curator note |

## Test Plan
- Visual inspection confirms valid INI syntax (sample file, not exercised by CI)
- `grep -n "network-online" scripts/claude-monitor.service` — no matches
- No Swift/build changes; this is a config/doc-only change to a sample systemd unit

Closes #14

### DIFF
--- a/scripts/claude-monitor.service
+++ b/scripts/claude-monitor.service
@@ -11,8 +11,10 @@
 
 [Unit]
 Description=Claude Monitor headless usage poller
-After=network-online.target
-Wants=network-online.target
+# No network-online ordering: this is a *user* unit (systemctl --user), and
+# network-online.target only has a provider in the system manager — the
+# ordering would be a silent no-op here. The poller retries transient network
+# errors itself, and Restart=on-failure covers startup before network is up.
 
 [Service]
 ExecStart=/usr/local/bin/claude-monitor


rjwalters:
Reviewed the diff for issue #14.

**Code review:** The change is correct and satisfies both acceptance criteria:
- Removes the no-op `After=network-online.target` / `Wants=network-online.target` lines from the `[Unit]` section (systemd user manager has no provider for `network-online.target`, so these lines were silently ignored — this diagnosis matches issue #14).
- Replaces them with a clear comment explaining *why* they were dropped (user unit vs. system unit distinction, plus the existing `Restart=on-failure` / poller retry logic covering the startup-before-network case), which should prevent someone re-adding them later without understanding the tradeoff.
- Resulting `[Unit]` section is valid systemd INI syntax; comment lines are permitted inside any section.
- Scope is appropriately minimal — no unrelated changes, README correctly left untouched (it never referenced `network-online.target`).

**CI status:** `gh pr checks 17` currently shows the `build` check as `pending` (still in progress), and `mergeStateStatus` is `UNSTABLE` as a result. Per review policy, I can't approve on a pending/unresolved CI signal.

Not applying an end-state label yet — leaving `loom:review-requested` in place so this gets re-evaluated once the `build` check finishes. Code itself looks ready to merge assuming CI comes back green.

rjwalters:
✅ **Approved!**

CI (`build`) now passes and `mergeStateStatus` is `CLEAN`. Combined with the code review above (correct fix, satisfies both acceptance criteria from #14, minimal well-scoped diff, clear rationale comment preventing regression), this is ready to merge.

PR #23, opened 2026-07-30
fix: backfill accounts.email from a well-formed account_name label

rjwalters:
## Summary

Accounts could persist indefinitely with `email = NULL` even when their
`account_name` was itself a well-formed address. This broke `loom-daemon
tokens import-from-monitor`, which joins accounts on `email` — newly-added
capacity became silently invisible to the fleet.

## Root Cause

Root cause traced to two paths that only ever set `account_name`, never
`email`, and no backfill/retry existed:

1. **Add-time**: pasting a bare token (`UsagePopoverView.addToken()` →
   `OAuthPoller.addAccountWithToken(token)`) has no known email, so `email`
   is inserted as `NULL`.
2. **Rename-time**: `UsageStore.updateAccountName()` lets a user set a custom
   label/alias — commonly the account's own email address, to tell accounts
   apart in the UI — but only ever updated `account_name`, never `email`.

Once an account reaches state 2 (label = a real address, `email` still
`NULL`), nothing in the codebase retries or backfills `email` — it stays
`NULL` forever, as observed for `agent-6/7/8@2amlogic.com`.

## Changes

- `OAuthPoller.swift`: add a shared `looksLikeEmailAddress(_:)` conservative
  well-formed-address check (distinguishes a real address from an opaque
  account-ID/UUID label). Use it in the account upsert: when the
  caller-supplied `email` is unavailable, fall back to the label if it looks
  like an address.
- `UsageStore.swift`:
  - `updateAccountName()`: when renaming to a well-formed address, backfill
    `email` too (never overwrites an existing `email`), and reflect the
    backfill immediately in in-memory state for UI feedback.
  - `ensureDatabase()`: add a one-time-per-launch healing migration
    (`backfillMissingEmailsFromAccountName`) that copies `account_name` into
    `email` for any existing account stuck with `email IS NULL` where the
    label is address-shaped — this heals accounts already broken by earlier
    app versions (the exact scenario reported in #15), not just future ones.
- `README.md`: document the `accounts` table contract — `email` is the
  stable join key external tooling (`loom-daemon tokens import-from-monitor`)
  should key on; `account_name` is a free-text, user-editable label with no
  guarantee of being an address.

| When the profile email is unavailable and the label is a well-formed address, fall back to the label for `email` | ✅ | `resolvedEmail = email ?? (looksLikeEmailAddress(label) ? label : nil)` in the upsert (`OAuthPoller.swift`); rename-time backfill in `updateAccountName()` |
| No account with valid credentials can persist indefinitely with `email = NULL` | ✅ | Add three defenses: (1) add-time fallback, (2) rename-time backfill, (3) one-time launch migration that heals *already*-NULL rows — verified against a scratch sqlite DB seeded with the exact reported scenario (`account_name = 'agent-8@2amlogic.com'`, `email = NULL`); the migration query heals it while leaving a genuinely opaque UUID label untouched |
| (Nice to have) document the `accounts` table contract | ✅ | Added to `README.md` "Database" section |

## Test Plan

- `swift build` — clean build, no new warnings introduced (two pre-existing
  unrelated warnings in `AnthropicAPI.swift` about an unused `data` binding,
  untouched by this change).
- Verified `looksLikeEmailAddress` logic in isolation against 11 cases
  (well-formed addresses, a raw UUID label, missing local/domain parts,
  double `@`, leading/trailing dots, embedded whitespace, empty string) — all
  passed.
- Verified the migration SQL against a throwaway sqlite3 DB seeded with three
  rows mirroring the issue's reported state (a broken `agent-8@...` row, an
  already-correct row, and an opaque-UUID-label row) — only the broken row
  got healed, matching expected behavior.
- Did not run the built binary against a real `~/.claude-monitor/usage.db`
  for this PR (macOS's `homeDirectoryForCurrentUser` ignores a `$HOME`
  override, so an isolated end-to-end run isn't possible without touching a
  real account database).

Closes #15

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,6 +403,19 @@ sqlite3 ~/.claude-monitor/usage.db \
   "SELECT timestamp, primary_percent FROM usage_history ORDER BY timestamp DESC LIMIT 10;"
 ```
 
+**`accounts` table contract for external consumers:** `email` is the stable
+join key external tooling should key off of — notably `loom-daemon tokens
+import-from-monitor`, which matches accounts by `email` to build its token
+pool. `account_name` is a free-text, user-editable display label/alias and is
+**not** guaranteed to be an address (though it often is, since renaming an
+account to its own email is a common way to tell accounts apart in the UI).
+The app backfills `email` from `account_name` whenever the profile-derived
+email is unavailable but the label is itself a well-formed address — both at
+add/rename time and via a one-time healing migration on launch — so no account
+with valid credentials should persist indefinitely with `email = NULL`. If you
+ever see one, it means `account_name` isn't address-shaped either; there's no
+address for external tooling to recover.
+
 ## Uninstall
 
 ```bash
--- a/menubar-app/ClaudeMonitor/Sources/OAuthPoller.swift
+++ b/menubar-app/ClaudeMonitor/Sources/OAuthPoller.swift
@@ -3,6 +3,25 @@ import Foundation
 private let flog = FileLogger.shared
 private let fcat = "OAuth"
 
+/// A conservative "well-formed enough" email check — not full RFC 5322
+/// validation, just enough to distinguish a real address (e.g. one a user
+/// typed as an account label/alias) from an opaque account ID (a UUID/org ID,
+/// which never contains '@'). Used to backfill the `accounts.email` column —
+/// the join key `loom-daemon tokens import-from-monitor` uses to find
+/// accounts (#15) — whenever a profile-fetch-derived email is unavailable but
+/// the label plainly carries the address.
+func looksLikeEmailAddress(_ s: String) -> Bool {
+    guard !s.isEmpty, !s.contains(where: { $0.isWhitespace }) else { return false }
+    let parts = s.split(separator: "@", omittingEmptySubsequences: false)
+    guard parts.count == 2 else { return false }  // exactly one '@'
+    let local = parts[0]
+    let domain = parts[1]
+    guard !local.isEmpty, domain.contains("."), !domain.hasPrefix("."), !domain.hasSuffix(".") else {
+        return false
+    }
+    return true
+}
+
 // MARK: - Token Status
 
 enum TokenStatus: String {
@@ -275,6 +294,13 @@ class OAuthPoller: ObservableObject {
             let now = ISO8601DateFormatter().string(from: Date())
             let label = orgName ?? email ?? accountId
 
+            // When the profile/caller-supplied email is unavailable, fall back to
+            // the label itself if it's a well-formed address — an account must
+            // never persist indefinitely with email = NULL just because its
+            // profile fetch failed (or was never attempted) while the label
+            // plainly carries the address (#15).
+            let resolvedEmail = email ?? (looksLikeEmailAddress(label) ? label : nil)
+
             // Upsert account — never overwrite account_name (user may have renamed)
             try db.run("""
                 INSERT INTO accounts (id, account_name, email, plan, last_updated, sort_order)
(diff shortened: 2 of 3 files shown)


rjwalters:
**Review: Approved**

Verified against the actual diff (not just the PR summary), plus a build in the checked-out worktree.

**Correctness**

- `looksLikeEmailAddress(_:)` (OAuthPoller.swift): conservative and sound — rejects empty/whitespace strings, requires exactly one `@`, non-empty local part, and a dotted domain that doesn't start/end with `.`. Verified it correctly rejects UUID-shaped org/account IDs (no `@`) and multi-`@` strings, and accepts well-formed addresses. No false-positive risk I could find.
- `updateAccountName()` rename-time backfill: uses `UPDATE ... SET email = COALESCE(email, ?)` and then re-reads the actual persisted value for the in-memory model — this is correct and provably never overwrites an existing email, only fills a NULL. This is the fix that resolves the exact scenario in #15 (label renamed to a real address, `email` column never touched).
- `backfillMissingEmailsFromAccountName()` migration: scoped by `WHERE email IS NULL AND account_name IS NOT NULL` in the SELECT, and re-guarded by `AND email IS NULL` in the UPDATE. Idempotent (a healed row no longer matches the SELECT, so reruns are a cheap no-op) and correctly scoped — it can't touch a row that already has a real email or a non-address-shaped label. Runs inside `ensureDatabase()`, which I confirmed is invoked from all three entry points (`main.swift`, `HeadlessRunner.swift`, and the popover add/import flows), so headless/Linux hosts get the same healing as the macOS UI.

**One non-blocking observation on the OAuthPoller.swift fallback:** `resolvedEmail = email ?? (looksLikeEmailAddress(label) ? label : nil)` is only reachable when the `email` param is `nil`, and in that branch `label = orgName ?? accountId` — but both current call sites of `saveCredentialForAccount` (from `addAccountWithToken`) hardcode `orgName: nil`, so `label` reduces to the org UUID, which will essentially never satisfy `looksLikeEmailAddress`. In today's codebase this specific fallback path is correct but effectively inert (a no-op safety net for a future caller that passes an email-shaped `orgName`) — it doesn't materially change add-time behavior. That's fine: the acceptance criterion ("no account persists indefinitely with `email = NULL`") is still fully satisfied by the rename-time backfill + one-time healing migration, which are the mechanisms that actually cover the reported bug (#15's `agent-6/7/8@2amlogic.com` scenario is a rename-time case). Just flagging so the PR description's "three defenses" framing isn't taken to mean layer 1 is presently doing work — worth a follow-up someday if a caller ever supplies a real `orgName`, but not worth blocking on.

**On the flagged `$HOME`/real-DB touch:** confirmed `UsageStore.init()` computes `dbPath` via `FileManager.default.homeDirectoryForCurrentUser` (pre-existing code, untouched by this PR) — this API is known to ignore a `$HOME` env override on macOS, unlike `NSHomeDirectory()`. That's a genuine macOS API quirk affecting test isolation, not something introduced here, and not evidence of a production bug (in production it's supposed to resolve the real user's home dir). Since the migration only heals rows matching `email IS NULL` + address-shaped `account_name`, a stray `--headless --once` run against the real `usage.db` would at worst have healed already-broken rows (a good outcome) or been a no-op read/poll cycle — consistent with "harmless." Recommend future manual testing use a throwaway DB path directly (if/when the app grows a `--db-path` override) rather than relying on `$HOME`, but this is a process note, not a PR blocker.

**Other checks**
- `swift build` succeeds clean in the checked-out worktree (`.loom/worktrees/issue-15`).
- CI (`build`) passes; `mergeStateStatus` is `CLEAN`.
- No test target exists in this SPM package (executable target only, no XCTest) — consistent with the rest of the repo's conventions; the builder's scratch-sqlite3 validation of the migration SQL and the 11-case manual check of `looksLikeEmailAddress` are a reasonable substitute given that.
- README addition documenting the `accounts` table contract (email vs. account_name) is accurate and directly useful for `loom-daemon tokens import-from-monitor`.
- PR body correctly uses `Closes #15`.

Nice, precise fix — approving.

Issue #25, opened 2026-07-30
Epic: multi-provider support — monitor OpenAI GPT (Codex) subscription accounts alongside Claude


rjwalters:
## Goal

Extend the monitor beyond Anthropic: track ChatGPT subscription accounts (as used by OpenAI's Codex CLI) with the same session %, weekly %, resets, and headroom model the app already has. This feeds directly into Loom's multi-runtime roadmap (rjwalters/loom#4489 "provider-aware account management") — the ranking the app exports must eventually let loom-daemon pick the right provider's token for the right runtime.

Decided 2026-07-30: provider support ships **first, under the current name**. A rebrand (e.g. "agent-monitor") is explicitly **deferred** to a possible 2.0.0 after this proves out — `~/.claude-monitor/` is a shared contract with nine loom-daemon source files plus `LOOM_CLAUDE_MONITOR_DIR` / `import-from-monitor`, so renaming requires a coordinated loom release and data-dir migration, and is not worth blocking provider work on.

## Why the model transfers

ChatGPT/Codex subscription accounts authenticate via OAuth (access/refresh tokens in `~/.codex/auth.json`) and have the same two-window rate-limit shape (a ~5h window plus a weekly window). The popover columns, DB schema, ranking export, and headroom sort are already provider-agnostic in everything but name.

## Phases

1. **Discovery spike (#next)** — prove a cheap authenticated probe can read a ChatGPT subscription account's usage/limits (the `anthropic-ratelimit-unified-*` analog). Everything else is contingent on this.
2. **Provider abstraction** — `provider` column on `Account` (migration, default `anthropic`); `AnthropicAPI` behind a small provider protocol; per-provider ping paths in `OAuthPoller`; provider badge in the popover.
3. **OpenAI GA + ranking export** — additive `provider` field in `~/.claude-monitor/ranking.json`; document schema for loom consumers.
4. **Loom side (tracked upstream, not here)** — provider-aware token selection in loom-daemon (rjwalters/loom#4489); coordinate with loom#4492 (Codex account lifecycle CLI) rather than duplicating account lifecycle. Token rolling stays Anthropic-only initially.

## Out of scope

- Rebrand/rename (deferred, see above)
- OpenAI API-key (platform) accounts — this is about subscription plans, matching the app's existing focus
- Codex token rolling/lifecycle (loom#4492's territory)

rjwalters:
Decomposed into phase issues after spike #26 live-verified the OpenAI contract:

[...]

PR #30, opened 2026-07-30
feat: provider-agnostic rate-limit model + provider column on accounts

### LINKED ISSUE #28: [Epic #25 Phase 2] Provider abstraction: Account schema + provider-agnostic rate-limit model
Part of #25 (multi-provider epic). This is the **foundation phase** — no OpenAI network calls ship here; it makes the data model and client layer provider-agnostic so Phase 3 can plug in the OpenAI poller.

Grounded in the live-verified contract from spike #26 — see `docs/spikes/2026-07-30-codex-usage-probe.md` § "Live verification (AUTHORITATIVE)". Do not design against the static-analysis hypotheses earlier in that document; they were corrected.

## Scope

1. **`Account` gains provider identity + refresh-capable credentials.**
   - `provider` column (`anthropic` | `openai`), defaulting existing rows to `anthropic` via migration.
   - `refresh_token` and `token_expires_at` columns. **This is mandatory infrastructure, not a nicety:** OpenAI access tokens live ~10 days (vs. Anthropic's ~1 year), so a poller that assumes an indefinitely-valid bearer will silently start failing. Anthropic rows leave both null.
   - Follow the healing-migration precedent from #15/#23 (backfill + repair on launch) rather than requiring users to re-add accounts.

2. **A provider-agnostic rate-limit window type.**
   The two providers report the same information in different shapes:
   - Anthropic: `anthropic-ratelimit-unified-{5h,7d}-*` response headers.
   - OpenAI: `rate_limit.primary_window` / `.secondary_window`, each `{used_percent, limit_window_seconds, reset_at}`.
   Introduce a shared window type carrying at minimum `usedPercent`, a window duration, and a reset instant, and have both providers populate it.
   - **Derive the window kind from its duration** (e.g. ~300s→session, ~604800s→weekly). Do **not** hard-code "primary = session, secondary = weekly": the verified probe returned a weekly `primary_window` with `secondary_window: null`.
   - **A nullable/missing session window must be a supported state** end-to-end (store, headroom score, popover). An OpenAI account may legitimately have no session figure to display.

3. **`AnthropicAPI` behind a provider protocol.**
   Extract the usage-fetch surface into a protocol; `AnthropicAPI` becomes one conformer. No behavior change for existing Anthropic accounts — this phase is a refactor plus schema work.

## Out of scope (Phase 3)

The OpenAI poller network path, the `/wham/usage` client, the token-refresh call against `auth.openai.com/oauth/token`, the `provider` field in `ranking.json`, and the popover provider badge.

## Acceptance criteria

- [ ] Migration adds `provider` / `refresh_token` / `token_expires_at`; existing accounts continue to work with no user action, verified by launching against a pre-migration `usage.db`.
- [ ] A shared rate-limit window type exists and Anthropic polling populates it; headroom score and popover read through it.
- [ ] A missing session window renders sensibly (no crash, no misleading 0%) — cover with a test or a documented manual check.
- [ ] `swift build` clean on macOS **and** the Linux headless target (portable core stays free of AppKit/SwiftUI/Combine/os.Logger per CLAUDE.md).
- [ ] No behavior change for Anthropic-only users.

## Affected Files

- `menubar-app/ClaudeMonitor/Sources/UsageStore.swift` — Account model, schema migration
- `menubar-app/ClaudeMonitor/Sources/AnthropicAPI.swift` — extract provider protocol
- `menubar-app/ClaudeMonitor/Sources/OAuthPoller.swift` — populate the shared window type
- `menubar-app/ClaudeMonitor/Sources/UsagePopoverView.swift` — tolerate a missing session window

rjwalters:
## Summary

Phase 2 of the multi-provider epic (#25): make the data model and client layer provider-agnostic so the OpenAI/Codex poller can plug in during phase 3. **No OpenAI network calls ship here** — this is a refactor plus schema work, designed against the live-verified contract in `docs/spikes/2026-07-30-codex-usage-probe.md` § "Live verification (AUTHORITATIVE)".

## Changes

**Shared rate-limit model** — new `Sources/RateLimitWindow.swift`
- `AccountProvider` (`anthropic` | `openai`) with a tolerant parse: absent, blank, or unrecognized values resolve to `.anthropic` so a strange row can never take the poll loop down.
- `RateLimitWindow` (usedPercent / durationSeconds / resetAt / status) and `RateLimitSnapshot` (optional session + optional weekly + named sub-limits).
- **Window kind is derived from the window's duration, never its position.** `RateLimitSnapshot(windows:)` files an unordered list into buckets, so OpenAI's `[primary_window, secondary_window]` lands correctly even though the verified probe returned a *weekly* `primary_window` with `secondary_window: null`.
- **Every window is optional**, end to end: NULL stays NULL in the DB, `headroomScore` returns nil (not a misleading 100), the popover renders "—".

**Provider protocol** — new `Sources/UsageProviderClient.swift`
- `UsageProviderClient` + `ProviderCredentials` / `ProviderUsageSnapshot` / `ProviderAPIError`.
- `AnthropicAPIError` becomes a typealias for `ProviderAPIError`, so every existing `catch AnthropicAPIError.unauthorized` site compiles unchanged.
- `AnthropicAPIClient` conforms as a pure adaptation layer over `pingToken` / `identifyToken`. `refreshCredentials` defaults to nil — correct for Anthropic's ~1yr tokens; OpenAI's ~10-day tokens override it in phase 3.

**Schema migration**
- `accounts.provider`, `oauth_credentials.provider`, `oauth_credentials.token_expires_at` (plus an idempotent `refresh_token` heal). Added with `ADD COLUMN ... NOT NULL DEFAULT 'anthropic'`, so existing rows backfill in place, plus a healing `UPDATE` for blank values — the #15/#23 heal-on-launch precedent, not a re-add-your-accounts migration.
- Credential columns live on `oauth_credentials`, **not** `accounts`: that is already the table holding token material, and `accounts` is read directly by `RankingExporter`, which must never see a secret.
- `token_expires_at` is ISO 8601 (matching `token_rolled_at`), deliberately distinct from the vestigial epoch-ms `expires_at` column this app has never written.
- Schema/migration extracted into `UsageStore.applySchema(_:)` so `AccountSync.importBundle` and the self-test can run it against an arbitrary database.
- Read paths (`loadFromDatabase`, `loadActiveCredentials`, `AccountSync.exportBundle`) select the new columns only when present, so a database opened before migration still loads.
- `AccountSync` export/import round-trips provider + token expiry; both fields are optional, so `formatVersion` stays 1 and old bundles still import (resolving to Anthropic).

**Reading through the model**
- `UsageRecord.rateLimit` exposes a stored reading as a `RateLimitSnapshot`; `headroomScore` moved from the macOS-only popover into the portable core and now delegates to it, so the headless Linux daemon scores accounts identically.
- Popover percent cells, reset cells, and the session/weekly/reset sort keys all read through the window model.

**Testing** — new `ClaudeMonitor selftest` subcommand
There is no XCTest target (the package has zero dependencies by design), so tests ship as a portable, network-free, credential-free subcommand that exits non-zero on failure and is wired into CI on **both** macOS and Linux.

- [x] **Migration adds `provider` / `refresh_token` / `token_expires_at`; existing accounts continue to work with no user action, verified by launching against a pre-migration `usage.db`.** Verified two ways: (1) the self-test synthesizes the exact v1.17.0 pre-#28 schema, populates it, migrates, and asserts column presence, `anthropic` backfill, idempotency, and that the account still loads; (2) `selftest --db <copy>` was run against a copy of the live 12-account `~/.claude-monitor/usage.db` — all 12 migrated, every row backfilled to `anthropic`, per-account headroom scores byte-identical before and after.
- [x] **A shared rate-limit window type exists and Anthropic polling populates it; headroom score and popover read through it.** `PingResponse.rateLimit` builds the snapshot; `writePingToDB`, `headroomScore`, `sortedAccountsForPopover`, `UsageStore.resetSeconds`, and every popover percent/reset cell and sort key read through it.
- [x] **A missing session window renders sensibly (no crash, no misleading 0%).** Covered by `testMissingSessionWindow` and `testSnapshotFromPositionalWindows`: a weekly-only reading scores off the weekly window, a window-less reading yields `nil` headroom, and the popover's `percentText`/`resetLabel` render "—" for nil.
- [x] **`swift build` clean on macOS and the Linux headless target.** Both build clean (only the pre-existing `UpdateChecker` capture-`self` and unused-`data` warnings). Portable core is free of AppKit/SwiftUI/Combine/os.Logger — grepped; `UsagePopoverView.swift` remains the only touched file with `import SwiftUI`, still inside its `#if os(macOS)` fence.
- [x] **No behavior change for Anthropic-only users.** `writePingToDB` keeps the `?? 0` coercion so stored values are unchanged; `sortedAccountsForPopover` keeps its `?? 0` comparison semantics; `AnthropicAPIError` is preserved as a typealias; the poll loop still calls `pingToken` directly.

## Test Plan

```
cd menubar-app/ClaudeMonitor
swift build && .build/debug/ClaudeMonitor selftest          # 44 checks pass (macOS)
cp ~/.claude-monitor/usage.db /tmp/ && .build/debug/ClaudeMonitor selftest --db /tmp/usage.db
                                                            # 62 checks pass, 12 real accounts migrated
docker run --rm -v "$PWD":/src -w /src swift:6.1 bash -c \
  'apt-get update -qq && apt-get install -y -qq libsqlite3-dev && \
   swift build --scratch-path /tmp/build && /tmp/build/debug/ClaudeMonitor selftest && \
   /tmp/build/debug/ClaudeMonitor --once'                   # Linux: build + 44 checks + smoke run
```

The self-test was also confirmed to *fail* correctly: inverting the session/weekly duration boundary produced 4 failures and exit 1.

## Notes for the reviewer

- **Slight scope extension:** `AccountSync` export/import was updated to carry `provider` / `tokenExpiresAt`. Without it, multi-host sync would silently reset an OpenAI account to `anthropic` on import once phase 3 lands. Fields are optional, so v1 bundles are unaffected.
- **One rendering nuance:** the popover's reset cell previously echoed an unparseable reset string verbatim; it now shows "—". Only ISO 8601 values are written by this app, so this is unreachable in practice.
- Out of scope per the issue and untouched: the OpenAI network path, `/wham/usage` client, `auth.openai.com/oauth/token` refresh call, `provider` in `ranking.json`, and the popover provider badge.

Closes #28

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,6 +17,11 @@ jobs:
           cd menubar-app/ClaudeMonitor
           swift build
 
+      - name: Self-test (portable core)
+        run: |
+          cd menubar-app/ClaudeMonitor
+          .build/debug/ClaudeMonitor selftest
+
   build-linux:
     runs-on: ubuntu-latest
     container: swift:6.1
@@ -30,6 +35,11 @@ jobs:
         run: |
           cd menubar-app/ClaudeMonitor
           swift build
+
+      - name: Self-test (portable core)
+        run: |
+          cd menubar-app/ClaudeMonitor
+          .build/debug/ClaudeMonitor selftest
 
       - name: Smoke run (no credentials)
         run: |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,7 +17,9 @@
 
 - The same package builds on Linux (`swift build` with `libsqlite3-dev` installed) as a headless daemon for Loom hosts — no UI, same poll loop, same `usage.db`/`ranking.json` outputs. See README "Headless Mode / Linux".
 - UI sources are fenced with `#if os(macOS)`; portable core must stay free of AppKit/SwiftUI/Combine/os.Logger. On Linux, `LinuxCompat.swift` shims `ObservableObject`/`@Published`, and `CSQLite/` maps the system libsqlite3 (macOS uses the SDK's `SQLite3` module).
-- Entry points: `main.swift` (macOS, dispatches to `HeadlessRunner` on `--headless`) and `HeadlessMain.swift` (Linux, always headless). Flags: `--once`, `--interval <sec>`, `--version`.
+- Entry points: `main.swift` (macOS, dispatches to `HeadlessRunner` on `--headless`) and `HeadlessMain.swift` (Linux, always headless). Flags: `--once`, `--interval <sec>`, `--version`. Subcommands: `accounts`, `selftest`.
+- Run the test suite with `swift build && .build/debug/ClaudeMonitor selftest` (exits non-zero on failure). CI runs it on macOS and Linux.
+- To verify a schema migration against real data, copy the live DB first — `cp ~/.claude-monitor/usage.db /tmp/ && .build/debug/ClaudeMonitor selftest --db /tmp/usage.db`. `--db` **writes** to the path it is given; never point it at `~/.claude-monitor/usage.db`.
 - Parse response headers via `extractAnthropicHeaders` (lowercased map), never `allHeaderFields` subscripts — those are case-sensitive on Linux.
 - Verify Linux builds from macOS with the `swift:6.1` Docker image (`apt-get install libsqlite3-dev`, then `swift build`).
 
@@ -29,8 +31,11 @@
   - `Sources/HeadlessMain.swift` / `Sources/HeadlessRunner.swift` - Linux entry point + UI-less poll loop (also `--headless` on macOS)
   - `Sources/LinuxCompat.swift` - `ObservableObject`/`@Published` stand-ins for Linux (no Combine)
   - `CSQLite/` - System-library target mapping Linux libsqlite3
-  - `Sources/AnthropicAPI.swift` - API client (usage, profile, token refresh)
+  - `Sources/RateLimitWindow.swift` - Provider-agnostic rate-limit model (`AccountProvider`, `RateLimitWindow`, `RateLimitSnapshot`). Window kind is **derived from the window's duration**, never its position in the provider response, and every window is optional — an account may legitimately report no session window.
+  - `Sources/UsageProviderClient.swift` - `UsageProviderClient` protocol + `ProviderCredentials` / `ProviderUsageSnapshot` / `ProviderAPIError` (aliased as `AnthropicAPIError`)
+  - `Sources/AnthropicAPI.swift` - API client (usage, profile, token refresh); conforms to `UsageProviderClient`
   - `Sources/OAuthPoller.swift` - Ping-based usage polling, token add/import, Fable probes
+  - `Sources/SelfTest.swift` - `ClaudeMonitor selftest`: portable-core assertions (window model, schema migration) with no network/credentials. Run in CI on both macOS and Linux; there is no XCTest target because the package has zero dependencies.
   - `Sources/UsagePopoverView.swift` - Summary-table popover, sortable headers, add-account dialog
   - `Sources/UsageChartView.swift` - Per-account usage-history chart window
   - `Sources/RollTokenView.swift` / `Sources/TokenRoller.swift` - Roll Token wizard + revoke-script generator. The workflow depends on undocumented, web-session-cookie-authenticated `claude.ai/api/oauth` endpoints; if they change, `TokenRoller.revokeAllScript` is the single place to update (see README "Rolling a Token").
--- a/README.md
+++ b/README.md
@@ -314,6 +314,7 @@ import — see [Multiple Accounts](#multiple-accounts)), then:
 claude-monitor                  # poll loop, logs to stdout + ~/.claude-monitor/debug.log
 claude-monitor --once           # one poll cycle, write ranking.json, exit
 claude-monitor --interval 300   # override per-account poll interval (seconds, min 60)
+claude-monitor selftest         # self-check (no network/credentials); non-zero exit on failure
 ```
 
 Edits to `accounts.env` / `accounts.local.env` are picked up automatically
--- a/menubar-app/ClaudeMonitor/Sources/AccountSync.swift
+++ b/menubar-app/ClaudeMonitor/Sources/AccountSync.swift
@@ -15,12 +15,17 @@ import Foundation
 enum AccountSync {
     static let formatVersion = 1
 
+    /// New optional fields (`provider`, `tokenExpiresAt`) decode as nil from a
+    /// pre-multi-provider bundle and resolve to Anthropic on import, so
+    /// `formatVersion` stays at 1 — old and new hosts still interoperate.
     struct ExportedCredential: Codable {
         var label: String
         var source: String
+        var provider: String?
         var accessToken: String?
         var refreshToken: String?
         var expiresAt: Int64?
+        var tokenExpiresAt: String?
         var scopes: String?
         var subscriptionType: String?
         var rateLimitTier: String?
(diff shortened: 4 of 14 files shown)


rjwalters:
## Review: Approved

Checked out `pr-30` (e02a2e0) and verified the following.

**Build**
- `swift build` clean on macOS from a fresh `.build` (only 4 instances of the pre-existing `unused-data` warning, no new warnings).
- CI: both `build-macos` and `build-linux` pass (Linux job was pending at review start, re-checked after and it's green).
- `mergeStateStatus` is `CLEAN`, `mergeable` is `MERGEABLE`.

**Self-test / migration verification**
- `ClaudeMonitor selftest` → 44/44 checks pass locally, matching the PR description.
- Ran `selftest --db <copy>` against a copy of my real 12-account `~/.claude-monitor/usage.db` (not the same copy the builder used, but same idea): 62/62 checks pass, migrated all 12 accounts, backfilled every row to `provider = anthropic`, and headroom scores were confirmed unchanged pre/post migration by the test's own before/after comparison. This corroborates the "no user action, no behavior change" claim in the PR body.

**Design/code review**
- `RateLimitWindow.swift`: window kind derived strictly from duration (`kind(forDuration:)`), never position — matches the acceptance criterion about OpenAI's weekly window potentially arriving in the "primary" slot. `RateLimitSnapshot(windows:)` correctly buckets by kind regardless of array order, and picks the more-used window on a kind collision. Every window is optional end-to-end; verified `headroomScore` returns `nil` (not 0 or 100) when nothing is known, and the popover renders "—" for exactly that case (`UsagePopoverView.swift`).
- Credential placement: confirmed `RankingExporter.swift` only ever selects `id, email, plan` (and `is_active`) from `accounts`/`oauth_credentials` — never a token column — so putting `refresh_token`/`token_expires_at` on `oauth_credentials` rather than `accounts` is the right call and keeps the "no secrets" invariant intact.
- Migration: `ADD COLUMN ... DEFAULT 'anthropic'` + idempotent healing `UPDATE`, matches the established #15/#23 heal-on-launch precedent. `applySchema` is correctly reused by both the app's normal startup path and `AccountSync.importBundle`, so a never-launched host importing a bundle still gets the new columns before the upserts run.
- No-behavior-change claim checked directly: `saveCredentialForAccount`'s new `provider`/`refreshToken`/`tokenExpiresAt` parameters all default appropriately, and both existing call sites in `addAccountWithToken` don't pass them — so existing Anthropic flows are provably untouched. `writePingToDB` keeps the `?? 0` coercion, so stored values for existing users are unchanged.
- `AnthropicAPIError` → `ProviderAPIError` typealias: grepped for `AnthropicAPIError` usages elsewhere in the tree, all still compile against the alias as claimed.
- Read paths (`UsageStore.loadFromDatabase`, `OAuthPoller.loadActiveCredentials`, `AccountSync.exportBundle`) all gate the new columns on `tableColumns(db, ...).contains(...)`, so a pre-migration DB opened by an external tool still loads without throwing.
- Portable core stays free of AppKit/SwiftUI/Combine/os.Logger — `RateLimitWindow.swift`, `UsageProviderClient.swift`, and `SelfTest.swift` are all plain Foundation, and `UsagePopoverView.swift` remains the only touched file with `import SwiftUI`, correctly inside its `#if os(macOS)` fence.
- Documentation (`CLAUDE.md`, `README.md`) updated consistently with the new files/subcommand.

**Minor, non-blocking nit**
- `UsageStore.resetSeconds`'s new `for date in [...].compactMap({ $0 }) { return date... }` is a slightly unusual way to express "return the first non-nil date" (a `.first` would read more directly) — purely stylistic, not a bug, and behavior matches the original priority order (session before weekly).

All five acceptance criteria in #28 are met and independently verified rather than just taken on the builder's word. No security issues found. Approving.

Issue #26, opened 2026-07-30
spike: can a cheap authenticated probe read ChatGPT/Codex subscription usage + rate-limit windows?


rjwalters:
Part of #25 (multi-provider epic). This spike gates everything else in the epic — do not start the provider abstraction until this answers yes/no.

## Question

The app's Anthropic polling sends a 1-token Haiku message and reads `anthropic-ratelimit-unified-*` response headers. Is there an OpenAI equivalent for **ChatGPT subscription accounts** (Plus/Pro, as used by Codex CLI): a minimal authenticated request whose response carries session-window and weekly-window usage/reset data?

## Approach

1. Authenticate with an existing Codex CLI credential (`~/.codex/auth.json` — access/refresh OAuth tokens). Use a low-value/test account, NOT a primary account.
2. Observe what Codex CLI itself does: it displays 5h/weekly usage percentages, so the signal exists somewhere — capture its traffic (mitmproxy or `--verbose`-style logging if available) to find the endpoint and response fields.
3. Prototype the probe as a scratch script (curl or Swift): minimal request, dump status + headers + relevant body fields.
4. Document: endpoint, auth header shape, usage/reset fields, cost per probe (tokens/requests consumed), refresh-token dance if the access token is short-lived (unlike Anthropic's ~1yr `sk-ant-oat01` tokens — this difference may drive the Account schema).

## Acceptance criteria

- [ ] A documented, reproducible probe (script checked into the issue or a draft PR) that returns current session-window and weekly-window usage for a ChatGPT subscription account, or a clear write-up of why it isn't feasible.
- [ ] Notes on token lifetime/refresh requirements vs. the app's current long-lived-token assumption.
- [ ] Recommendation: proceed to epic phase 2, or stop.

## Notes

- Undocumented endpoints are acceptable — the Roll Token workflow already depends on undocumented `claude.ai/api/oauth` endpoints, with the fragility documented in README. Same convention applies here.
- Credential handling: never print live tokens; test against scratch credentials; see the #16 build's transcript-leak incident for the failure mode to avoid (`FileManager.homeDirectoryForCurrentUser` ignores `HOME` overrides on macOS — use explicit paths).

Issue #16, opened 2026-07-30
feature: headless CLI to export/import account records + credentials for multi-host sync


rjwalters:
## Context

The fleet now runs claude-monitor-backed token pools on multiple hosts (two Macs + Linux cloud workers). When accounts are added on one host, propagating them today is entirely manual — 2026-07-30's relay of three new accounts took: sqlite surgery on the source DB (see the email=NULL bug), `loom-daemon tokens import-from-monitor` on the source host, then `scp` of raw token files plus a hand-merge into each destination pool's `index.json`. Worse, the *destination hosts' monitor DBs stay stale* — only the loom pools were synced, so each host's monitor UI/rankings disagree about which accounts exist.

## Ask

A programmatic, headless mechanism to converge account records across hosts. Shape (suggestion, not prescription):

```
claude-monitor accounts export [--json] > accounts.json     # records + credentials (encrypted or plaintext-with-warning)
claude-monitor accounts import accounts.json                # upsert by email; never clobbers newer last_updated
```

- Headless-safe (Linux workers have no menubar app — the CLI must not require the GUI).
- Upsert semantics keyed on email, preserving local usage history/rankings (sync identities+credentials, not usage).
- Sane credential handling for transport (age/openssl encryption, or at minimum 0600 output + a loud warning).
- With this, a fleet relay becomes: export on the source → push to N hosts → import → `loom-daemon tokens import-from-monitor` on each (which then has a correct local source of truth), replacing the scp-token-files-and-patch-index dance.

## Suggested acceptance criteria

- [ ] Export/import round-trips an account (record + oauth credential) between two hosts headlessly.
- [ ] Import is idempotent and never regresses a newer local record.
- [ ] Documented as the blessed multi-host propagation path (README section).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #31, opened 2026-07-30
feat: poll OpenAI/Codex usage and tag ranking.json with provider

### LINKED ISSUE #29: [Epic #25 Phase 3] OpenAI usage poller + provider-aware ranking export
Part of #25 (multi-provider epic). Depends on #28 (Phase 2 provider abstraction must land first — this phase plugs into the protocol and schema it introduces).

Ships the actual OpenAI/Codex polling path on top of Phase 2's abstraction, using the live-verified contract in `docs/spikes/2026-07-30-codex-usage-probe.md` § "Live verification (AUTHORITATIVE)".

## Verified contract (do not re-derive)

```
GET https://chatgpt.com/backend-api/wham/usage
Authorization: Bearer <access_token>
```

- Returns 200 with `rate_limit.primary_window` / `.secondary_window`, each `{used_percent, limit_window_seconds, reset_after_seconds, reset_at}`, plus `allowed` / `limit_reached` booleans.
- Identity (`email`, `plan_type`, `account_id`) arrives in the **same** response — no separate profile call needed.
- Per-model sub-limits arrive in `additional_rate_limits[]` (`limit_name`, `metered_feature`, same nested `rate_limit` shape).
- **The `/backend-api/codex/usage` route is Cloudflare-challenged (403) — do not use it.**
- No `chatgpt-account-id` header is required; the bearer token carries identity.

## Scope

1. **OpenAI provider client** conforming to Phase 2's protocol: call `/wham/usage`, map the response onto the shared window type (deriving window kind from `limit_window_seconds`, tolerating a null `secondary_window`).
2. **Token refresh.** `POST https://auth.openai.com/oauth/token` with the stored `refresh_token`, invoked proactively before `token_expires_at` rather than reactively on 401. **This is the riskiest unverified piece** — the spike confirmed the endpoint from Codex CLI's own login log but never exercised the call. Verify it works before wiring it into the poll loop, and fail loudly (surface a token-health state in the popover) rather than silently polling with an expired token.
3. **Credential import.** A path to get a Codex credential into the app — reading `~/.codex/auth.json` (honoring `$CODEX_HOME`) is the obvious source, mirroring how bulk import reads `accounts.env`.
4. **`ranking.json` gains a `provider` field** (additive; consumers that ignore it keep working). Document the schema change for loom-daemon, which consumes this file for token selection.
5. **Provider badge** in the popover so a row's provider is visible at a glance.

## Consider surfacing per-model sub-limits

`additional_rate_limits[]` is a close analog of this app's existing per-model Fable tracking and is better-structured than anything Anthropic exposes. Wiring it into the per-account chart is in scope if it fits cleanly; punt to a follow-up if it balloons.

## Security constraints (non-negotiable)

- The response body contains PII (`email`, `account_id`, `user_id`) and the request carries a live bearer token. **Never log or print token values**; redact PII in any debug output, including nested fields (the #26 spike's near-miss was a nested `email` claim that a top-level-only redaction pass missed).
- Test against explicit scratch paths, never `$HOME`-relative defaults — `FileManager.homeDirectoryForCurrentUser` ignores `HOME` overrides on macOS, which is exactly how #16 leaked live tokens into a transcript.

## Acceptance criteria

- [ ] An OpenAI account can be added and shows live usage in the popover alongside Anthropic accounts.
- [ ] Token refresh is exercised and verified against the real endpoint; an expired/unrefreshable token surfaces as a visible token-health state, not a silent poll failure.
- [ ] `ranking.json` carries `provider`; existing loom consumers are unaffected by the addition.
- [ ] A null `secondary_window` renders correctly (no fabricated session %).
- [ ] `swift build` clean on macOS and Linux headless.

## Affected Files

- `menubar-app/ClaudeMonitor/Sources/OpenAIAPI.swift` — new provider client
- `menubar-app/ClaudeMonitor/Sources/OAuthPoller.swift` — per-provider poll paths, refresh scheduling
- `menubar-app/ClaudeMonitor/Sources/RankingExporter.swift` — additive `provider` field
- `menubar-app/ClaudeMonitor/Sources/UsagePopoverView.swift` — provider badge
- `README.md` — document adding an OpenAI account

rjwalters:
## Summary

Ships the OpenAI/ChatGPT (Codex) polling path on top of phase 2's provider
abstraction, using the live-verified contract in
`docs/spikes/2026-07-30-codex-usage-probe.md` § "Live verification". OpenAI
accounts now sit in the same table as Anthropic ones, are polled on the same
cadence, and appear in `ranking.json` tagged with their provider.

The two findings that broke the original static analysis are honored
structurally, not just documented: window kind is derived from
`limit_window_seconds` (never from `primary`/`secondary` position), and a null
`secondary_window` produces *no* session window rather than a 0% one.

## Changes

**New — `Sources/OpenAIAPI.swift`**
- `OpenAIAPIClient` conforming to `UsageProviderClient`. `GET
  chatgpt.com/backend-api/wham/usage` (never `/backend-api/codex/usage`, which
  is Cloudflare-challenged) mapped onto the shared `RateLimitSnapshot`.
  Per-model `additional_rate_limits[]` land in the snapshot's `named` map;
  identity (`email` / `plan_type` / `account_id`) is read from the same
  response, so there is no separate profile call.
- Token refresh via `POST auth.openai.com/oauth/token`, with `expires_in` or
  the access token's own `exp` claim as the expiry source (~10 days).
- `CodexAuth` reads Codex CLI's `~/.codex/auth.json`, honoring `$CODEX_HOME`.

**New — `Sources/CodexCLI.swift`**: `claude-monitor codex import [--auth <path>]
[--db <path>]`, a one-shot CLI (no `--headless` needed) that works on Linux.

**`Sources/OAuthPoller.swift`**
- Per-provider poll paths; a single provider-agnostic DB writer replaces the
  Anthropic-only one. Absent windows are stored as **NULL**, not 0, so a
  window-less provider can't inflate its apparent headroom. Anthropic's
  long-standing `?? 0` behavior is untouched.
- Proactive refresh 6h ahead of expiry, never reactive-on-401. Failure is loud:
  a new `CredentialExpiredError` keeps the `.expired` state (red dot) and its
  actionable message from being flattened to the generic "revoked" by the retry
  loop; a failed-but-still-valid renewal shows `.refreshing` (yellow) with the
  reason, and that warning survives a subsequent successful read.
- Fable probes, the `ACCOUNT_KEY_N` env export, and paste-replace semantics are
  now scoped to Anthropic accounts — an Anthropic-only paste can no longer
  delete OpenAI accounts the env format cannot express.

**`Sources/RankingExporter.swift`**: every account carries `provider`.
`schema` deliberately stays **1** (see Risks below). Adds an explicit
`exportNow(dbPath:outputPath:)` for scratch-path testing.

**`Sources/UsagePopoverView.swift`**: per-row `ProviderBadge` inside the
existing Account column (no column-width or popover-width changes); an "Import
Codex Account" button; Roll Token hidden for non-Anthropic rows.

**Docs**: README gains "Adding an OpenAI (Codex) Account" and a full
"Ranking Export (`ranking.json`)" section written for loom-daemon; the spike doc
records the refresh-endpoint probe result.

## Security

- No token value or raw response body is ever logged. Error paths surface only
  the OAuth error *code*.
- The archived field map redacts credential/PII keys **at every nesting depth**,
  which is the exact miss flagged in spike #26's process note. A self-test
  asserts on a deliberately nested `email`/`name` claim.
- Every test uses explicit scratch paths, never `$HOME`-relative defaults
  (`homeDirectoryForCurrentUser` ignores a `HOME` override on macOS — issue #16).
  Scratch data created during manual verification was deleted.

- [x] **An OpenAI account can be added and shows live usage alongside Anthropic
  accounts.** Verified end-to-end against the real credential into a *scratch*
  database seeded with a legacy (pre-migration) Anthropic account. Result: the
  legacy row migrated in place as `anthropic`, the new row landed as
  `provider=openai, plan=pro`, `usage_history` recorded `weekly=14%` with
  `session_percent = NULL`, and `oauth_credentials` stored the refresh token
  plus `token_expires_at = 2026-08-08T21:59:15Z`.
- [x] **Token refresh exercised against the real endpoint.** `POST
  auth.openai.com/oauth/token` was probed with the exact request the client
  sends and a deliberately invalid refresh token. It returned **401** with
  `{"error":{"code":"invalid_refresh_token", ...}}` — proving the endpoint, the
  JSON (not form-encoded) body, the `client_id`, and the `grant_type` are all
  accepted, with only the token itself rejected. It also revealed the error body
  is **nested**, not the flat RFC 6749 string, which the client's error
  extraction is written against and a self-test now pins.
  **A successful exchange was deliberately not run** — see Risks.
- [x] **An expired/unrefreshable token is a visible state, not a silent
  failure.** Verified with a synthetic already-expired `auth.json`: fails loudly
  (`Invalid OpenAI credential: Unauthorized — token may be expired or revoked`,
  exit 1) with no token material in the message. In the poll loop this is
  `.expired` (red) with the actionable message preserved.
- [x] **`ranking.json` carries `provider`; existing consumers unaffected.**
  Generated output: Anthropic row unchanged (`5h` + `7d`), OpenAI row
  `{"provider":"openai","utilization":{"7d":0.14}}` with **no `5h` key**,
  `schema` still 1, no credential material anywhere in the file.
- [x] **A null `secondary_window` renders correctly (no fabricated session %).**
  NULL in the DB, key omitted from `ranking.json`, `—` in the popover, and the
  headroom score computed from the weekly window alone (86).
- [x] **`swift build` clean on macOS and Linux headless.** Both verified; see
  Test Plan.

## Test Plan

- `swift build` on macOS — clean (only pre-existing warnings).
- **Linux verified for real, not reasoned about**: `docker run swift:6.1` +
  `libsqlite3-dev`, `swift build` exit 0, `selftest` 99/99 passing, and
  `codex --help` working from the Linux binary. One genuine portability bug was
  caught this way and fixed: `CFGetTypeID`/`CFBooleanGetTypeID` don't exist on
  Linux, so the archive flattener now walks a `JSONValue` decoded by
  `JSONDecoder` instead (which also keeps booleans distinct from 0/1 portably).
- `ClaudeMonitor selftest` — **99 checks**, up from 60, on both platforms.
  New coverage: the recorded wire body (weekly-primary + null-secondary,
  reordered windows, `limit_reached` → `rejected`, missing `account_id` throws),
  nested-PII redaction, JWT `exp` parsing, both OAuth error-body shapes,
  `auth.json` parsing from a scratch path, and the `ranking.json` export
  (provider present, `5h` omitted, schema unchanged, no secrets).
- `ClaudeMonitor selftest --wire <path>` — new flag that decodes a captured
  `/wham/usage` body offline and prints only derived numbers. Run against the
  real captured (scrubbed) response: mapped to weekly 14% / session none /
  headroom 86 / named `GPT-5.3-Codex-Spark` / 37 archived fields, PII redacted.

## Risks / Notes

- **A successful token refresh was not executed.** OAuth refresh may rotate the
  refresh token, which would invalidate the copy in `~/.codex/auth.json` and
  break the operator's Codex CLI login. The negative probe establishes the whole
  request contract short of that, and the client handles both outcomes (keeps
  the stored refresh token when none comes back, replaces it when one does).
  Documented in the README and in the spike doc's "Still unverified".
- **`schema` stays 1 on purpose.** The version tracks breaking changes; bumping
  it risks a strict consumer rejecting the document over a purely additive key.
  The README spells out the contract for loom-daemon, including that a missing
  `utilization["5h"]` means *unknown*, not 0.
- **Punted (worth a follow-up issue):** charting `additional_rate_limits[]` per
  model in the per-account chart window. The data is captured end-to-end today —
  it reaches `RateLimitSnapshot.named` and the `probe_snapshots` archive — but
  surfacing it in `UsageChartView` needs its own history/storage shape, which
  would have doubled this diff. Nothing here blocks it.
- Unrelated pre-existing warnings in `AnthropicAPI.swift` / `UsageStore.swift`
  were left alone.

Closes #29

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -34,14 +34,16 @@
   - `Sources/RateLimitWindow.swift` - Provider-agnostic rate-limit model (`AccountProvider`, `RateLimitWindow`, `RateLimitSnapshot`). Window kind is **derived from the window's duration**, never its position in the provider response, and every window is optional — an account may legitimately report no session window.
   - `Sources/UsageProviderClient.swift` - `UsageProviderClient` protocol + `ProviderCredentials` / `ProviderUsageSnapshot` / `ProviderAPIError` (aliased as `AnthropicAPIError`)
   - `Sources/AnthropicAPI.swift` - API client (usage, profile, token refresh); conforms to `UsageProviderClient`
-  - `Sources/OAuthPoller.swift` - Ping-based usage polling, token add/import, Fable probes
-  - `Sources/SelfTest.swift` - `ClaudeMonitor selftest`: portable-core assertions (window model, schema migration) with no network/credentials. Run in CI on both macOS and Linux; there is no XCTest target because the package has zero dependencies.
+  - `Sources/OpenAIAPI.swift` - OpenAI/Codex client: `GET chatgpt.com/backend-api/wham/usage` (never `/backend-api/codex/usage` — Cloudflare-challenged), `POST auth.openai.com/oauth/token` refresh, and `CodexAuth` (reads `~/.codex/auth.json`, honoring `$CODEX_HOME`). Access tokens live ~10 days; expiry comes from the token's own `exp` claim. **Never log a token or a raw response body** — `OpenAIUsageResponse.flatten` redacts PII/credential keys at *every* nesting depth before anything is archived or logged.
+  - `Sources/CodexCLI.swift` - `claude-monitor codex import` (one-shot CLI, no `--headless` needed, works on Linux)
+  - `Sources/OAuthPoller.swift` - Per-provider usage polling (Anthropic ping / OpenAI usage GET), token add/import, proactive OpenAI token refresh, Fable probes (Anthropic only). Absent windows are stored as NULL, not 0, for non-Anthropic providers.
+  - `Sources/SelfTest.swift` - `ClaudeMonitor selftest`: portable-core assertions (window model, schema migration, OpenAI wire mapping + PII redaction, `ranking.json` provider field) with no network/credentials. Run in CI on both macOS and Linux; there is no XCTest target because the package has zero dependencies. `--wire <path>` decodes a captured `/wham/usage` body offline (prints only derived numbers, never identity).
+  - `Sources/RankingExporter.swift` - Emits `~/.claude-monitor/ranking.json`. Every account carries `provider`; `schema` stays **1** because the field is additive. A `provider: openai` account may omit `utilization["5h"]` — consumers must read a missing key as *unknown*, never 0.
   - `Sources/UsagePopoverView.swift` - Summary-table popover, sortable headers, add-account dialog
   - `Sources/UsageChartView.swift` - Per-account usage-history chart window
   - `Sources/RollTokenView.swift` / `Sources/TokenRoller.swift` - Roll Token wizard + revoke-script generator. The workflow depends on undocumented, web-session-cookie-authenticated `claude.ai/api/oauth` endpoints; if they change, `TokenRoller.revokeAllScript` is the single place to update (see README "Rolling a Token").
   - `Sources/UsageStore.swift` - SQLite-backed data store, account/usage models
   - `Sources/SQLiteDB.swift` - Minimal wrapper over system libsqlite3 (zero package deps)
-  - `Sources/RankingExporter.swift` - Emits `~/.claude-monitor/ranking.json`
   - `Sources/FileLogger.swift` - Debug log at `~/.claude-monitor/debug.log`
   - `Assets/` - App icon (`AppIcon.icns` + 1024px master PNG + `icon-master.source.json` generation recipe)
 - `scripts/build-macos-app.sh` - Build script that compiles and creates the .app bundle (bundles the icon)
--- a/README.md
+++ b/README.md
@@ -38,6 +38,9 @@ OAuth tokens you provide, and renders the data locally on your Mac.
 - **Multi-account.** Add accounts one-by-one via a token from
   `claude setup-token`, or bulk-import from a `.env` file with
   `ACCOUNT_EMAIL_N` / `ACCOUNT_KEY_N` pairs.
+- **Multi-provider.** Anthropic and OpenAI/ChatGPT (Codex) accounts sit side by
+  side in the same table, each row tagged with a provider badge. See
+  [Adding an OpenAI (Codex) Account](#adding-an-openai-codex-account).
 - **Roll Token wizard.** Guided revoke-all + re-mint for an account's
   long-lived token (right-click its row → "Roll Token…"). A temporary stopgap
   until Anthropic ships a token-management API — see
@@ -114,6 +117,52 @@ and **appends** any new emails), then imports the result. Loading is
 **additive**: accounts in the lists are added or have their token refreshed, but
 accounts already in the app that aren't listed are left untouched — nothing is
 removed. Store these files with `chmod 600`; they contain live tokens.
+
+### Adding an OpenAI (Codex) Account
+
+ChatGPT subscription accounts (Plus/Pro, the ones Codex CLI uses) are polled
+too. Their usage comes from a dedicated read-only endpoint —
+`GET https://chatgpt.com/backend-api/wham/usage`, the same one Codex CLI's own
+`/usage` command calls — so no inference request is burned to read it.
+
+A ChatGPT credential is not a single pasteable string (it's an access token +
+refresh token + expiry), so it's **imported** from Codex CLI's own credential
+store rather than typed in:
+
+```bash
+codex login                    # once, if you haven't already
+claude-monitor codex import    # or: Add Account → "Import Codex Account"
+```
+
+The importer reads `$CODEX_HOME/auth.json` when `CODEX_HOME` is set, otherwise
+`~/.codex/auth.json`. Pass `--auth <path>` to read a different file. The
+credential is validated against the live usage endpoint before it's stored, and
+the account's email, plan, and OpenAI account id all come back in that same
+response — there's no separate profile call.
+
+What differs from an Anthropic row once it's added:
+
+- **Session % may be blank.** OpenAI reports its windows as
+  `primary_window` / `secondary_window`, each carrying its own length, and a
+  ChatGPT Pro account can legitimately report only a **weekly** window. When
+  there's no session window, the Session cells show `—`. That is a real
+  "unknown", not 0% — the headroom score and sorting use whichever windows
+  actually exist.
+- **Fable Left / Extra are always `—`.** Those columns track Anthropic premium
+  tiers; the Fable probe is skipped for OpenAI accounts entirely.
+- **Tokens expire, and are renewed for you.** An OpenAI access token lives
+  about 10 days. The poller renews it from the stored refresh token
+  *proactively* — 6 hours ahead of expiry, never waiting for a 401 — and the
+  Token dot reports the outcome: green (healthy), **yellow** (renewal is
+  failing but the current token still works; hover for the reason), **red**
+  (expired and unrenewable — re-run `claude-monitor codex import`). A stale
+  OpenAI account never fails silently.
+
+> **Note on refresh-token rotation.** If OpenAI issues a new refresh token
+> during renewal, this app stores it in its own database; it does **not** write
+> back to `~/.codex/auth.json`. Should Codex CLI ever start reporting a bad
+> credential after this app has been running a while, re-run `codex login` and
+> then `claude-monitor codex import`.
 
 ### Rolling a Token (revoke + re-mint)
 
@@ -226,22 +275,28 @@ against `/v1/messages`.
 
 ```
 ┌──────────────────────────────────────────────────────────────────────────┐
-│  OAuth tokens                                                            │
-│    - `claude setup-token` (single, paste-in)                             │
-│    - `.env` bulk import (ACCOUNT_EMAIL_N / ACCOUNT_KEY_N)                │
+│  OAuth credentials                                                       │
+│    - `claude setup-token` (single, paste-in)            [anthropic]      │
+│    - `.env` bulk import (ACCOUNT_EMAIL_N / ACCOUNT_KEY_N) [anthropic]    │
+│    - `claude-monitor codex import` (~/.codex/auth.json)   [openai]       │
 └────────────────────────────────┬─────────────────────────────────────────┘
                                  │ stored in
                                  ▼
 ┌──────────────────────────────────────────────────────────────────────────┐
 │  SQLite — ~/.claude-monitor/usage.db                                     │
 │    accounts │ oauth_credentials │ usage_history │ settings               │
+│    (accounts.provider / oauth_credentials.provider tag the upstream)     │
 └────────────────────────────────┬─────────────────────────────────────────┘
                                  │ read by
                                  ▼
 ┌──────────────────────────────────────────────────────────────────────────┐
-│  OAuthPoller → Anthropic API                                             │
-│    - Each account pinged once per 10 min (staggered)                     │
-│    - Tokens are long-lived (1 yr); no refresh dance needed               │
+│  OAuthPoller — one client per provider, one shared window model          │
+│    anthropic: POST /v1/messages (1-token ping) → rate-limit headers      │
+│               tokens are long-lived (~1 yr); no refresh dance needed     │
+│    openai:    GET /backend-api/wham/usage → rate_limit.{primary,         │
+│               secondary}_window; access tokens live ~10 days and are     │
+│               refreshed proactively via auth.openai.com/oauth/token      │
+│    - Each account polled once per 10 min (staggered)                     │
 └────────────────────────────────┬─────────────────────────────────────────┘
                                  │ data drives
                                  ▼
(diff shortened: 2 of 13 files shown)


rjwalters:
Follow-up for the punted piece (per-model `additional_rate_limits[]` charting) is filed as #32. The data is already captured end to end by this PR — it reaches `RateLimitSnapshot.named` and the `probe_snapshots` archive — so #32 is purely about surfacing it.

rjwalters:
Fixed the `build-macos` failure. Both checks are green — PR is back to review-ready for Judge (it was dispatched to Doctor ahead of review because CI caught a hard build error first).

**Root cause:** `CodexCLI.runImport` bridged sync → async with a `DispatchSemaphore` and assigned the awaited result to a captured `var`, which strict concurrency checking rejects:

```
CodexCLI.swift:72:13: error: mutation of captured var 'result' in concurrently-executing code
```

**Fix** (`menubar-app/ClaudeMonitor/Sources/CodexCLI.swift`, the only file touched):
- Dropped the semaphore and moved the whole continuation inside the `Task`, which already ends the process itself — the same sync→async idiom `HeadlessRunner` uses. No value crosses the concurrency boundary, so there is no captured `var` left to mutate.
- Also froze the parsed `--db` option into a `let` (`storePath`) before the boundary. CI surfaced this as a second error once the first was fixed — `dbPath` was a `var` too, and merely *referencing* a captured `var` is an error on that toolchain, not just mutating one.

Behavior and exit codes are unchanged (invalid credential still exits 1 with the same message).

**On the platform divergence:** `swift build` passing is not sufficient evidence for this class of bug — the swift:6.1 Linux container and my local Swift 6.3 toolchain both accept the original code; only CI's older macOS Xcode errors. I verified with `swift build -Xswiftc -swift-version -Xswiftc 6` as a strictly-stronger proxy, confirming it errors on the original and reports zero `CodexCLI` diagnostics on the fix.

**Verified locally:** clean `swift build` on macOS (0 errors), `selftest` 99/99 passing, `swift build` in the `swift:6.1` Linux container, and `codex import` exercised end-to-end against the live endpoint with a bogus credential (async path runs, `dispatchMain()` does not hang, exit 1).

**Left alone (pre-existing, out of scope):** the `immutable value 'data' was never used` warnings at `AnthropicAPI.swift:109` and `:263`, and the Linux-only `captured var 'self'` warnings in `UsageStore.swift` around line 1029. All exist on `main` and none are errors.

No other instance of the captured-var-in-`Task` pattern exists in the package — `CodexCLI.swift:71` was the only `DispatchSemaphore` bridge, and the `Task` at `HeadlessRunner.swift:46` is pre-existing and already safe (it captures only `let`s and mutates nothing).

rjwalters:
## Approved

Phase 3 of epic #25. Reviewed against the **live-verified** contract in `docs/spikes/2026-07-30-codex-usage-probe.md` § "Live verification (AUTHORITATIVE)" — not the static-analysis hypotheses earlier in that file. Endpoint, method, auth, and response shape all match: `GET https://chatgpt.com/backend-api/wham/usage`, bearer-only (`OpenAIAPI.swift:305-310` explicitly documents *why* no `chatgpt-account-id` header is sent), and `/backend-api/codex/usage` is never constructed anywhere in the tree.

CI green on both `build-macos` (macos-14) and `build-linux` (swift:6.1), `mergeStateStatus` CLEAN. Both CI jobs run `selftest`, so Linux portability and the 99 assertions are verified by the authoritative gate, not just locally.

### 1. Null session window — verified end to end, no fabricated 0%

Window kind is derived from duration in exactly one place (`RateLimitWindow.kind(forDuration:)`), and the OpenAI client feeds both slots through it positionally-agnostically: `RateLimit.windows` returns `[primaryWindow, secondaryWindow].compactMap { ... }` unordered, and `RateLimitSnapshot(windows:)` files each by `.kind`. A weekly `primary_window` lands in `weekly`; a null `secondary_window` leaves `session` nil. `Window.asRateLimitWindow` returns nil (not 0) when `used_percent` is absent.

Traced the nil through every downstream consumer:

- **Store** — `writeSnapshotToDB` passes `windows.session?.usedPercent` (a `Double?`) into the new provider-agnostic `writeUsageToDB`. `SQLiteDB.bindValues` maps `nil` to `sqlite3_bind_null`, and `usage_history.session_percent` is a nullable `REAL`. The Anthropic `?? 0` coercion is confined to `writePingToDB` with a comment explaining why the historical series stays 0-filled — the right call, no NULLs injected into an existing series.
- **`primary_percent`** — now `[session, weekly].compactMap{}.max()`, so a window-less account writes NULL rather than 0. Reset-detection is correctly gated behind `if let weeklyPercent`, so a provider with no weekly window can't trip a spurious synthetic rollover row.
- **Headroom** — `RateLimitSnapshot.headroomScore` returns nil when no window is known and scores off the weekly window alone otherwise; `selftest` pins 86 for the weekly-only fixture and 70 for the paired case.
- **Popover** — renders through `usage?.rateLimit.session?.usedPercent` into `percentText(Double?)`; sorting is on the same optional. No `?? 0`.
- **`ranking.json`** — `buildAccount` uses `if let s = sessionPercent { utilization["5h"] = ... }`, so the key is omitted, and `testRankingExportCarriesProvider` asserts `openaiUtil?["5h"] == nil` explicitly.
- **Headless log line** — rewritten to `compactMap{}.max()` and prints `?` rather than `0%`.

Phase 2's nil-not-zero discipline is preserved, and the fixture in `SelfTest.swift` deliberately encodes the exact shape that broke the static analysis (weekly primary + null secondary), plus a reordered-windows variant proving the session window still files correctly from the *secondary* slot.

### 2. Token refresh — proactive, and loud on failure

`refreshIfNeeded` (`OAuthPoller.swift:849`) runs at the top of `pollOpenAI`, gated on `credentials.isExpiring(within: refreshLeadTime)` where `refreshLeadTime = 6h` — well clear of the 10-minute poll cadence, so a token cannot expire between two cycles. There is no reactive-on-401 refresh path anywhere; the 401 handler goes straight to a status update.

The three failure modes are distinct and all visible:

- past expiry + unrenewable/refresh-threw → `.expired` (red) with an actionable message, `throw CredentialExpiredError`, **no request issued** with a known-dead token;
- expiring but still valid + refresh failed → `RenewalOutcome.warning` surfaced as `.refreshing` (yellow) with the reason, and — importantly — `pollOpenAI` deliberately does *not* clear that warning on a successful read (`OAuthPoller.swift` "A successful read does NOT clear a refresh warning"), which is what makes the proactive design actually worth having;
- no refresh token stored → the message names the recovery command.

The new `catch is CredentialExpiredError { return }` arm sits between the transient-retry arm and the generic arm, so the `.expired` state and its message survive instead of being flattened to the generic `.revoked` — the exact regression that catch exists to prevent. `persistRefreshedCredential` replaces access + refresh together (`COALESCE(?, refresh_token)`), correct for a provider that may rotate.

The request contract is verified as far as it safely can be: the negative probe against `auth.openai.com/oauth/token` pinned the JSON (not form-encoded) body, the `client_id`, the `grant_type`, and — a genuinely useful find — that the error body is nested `error.code`, not flat RFC 6749. `oauthErrorCode` handles both shapes and `selftest` pins both. Not running a successful exchange (to avoid rotating the refresh token out from under the operator's Codex CLI login) is the right trade, and the README documents the recovery path.

### 3. Secrets / PII — clean, including the two prior incident classes

- **No token values logged.** Grepped every added `flog.*` / `print` / stderr write in the diff: the closest any comes is `"OpenAI token refresh rejected (\(status)/\(code))"`, where `code` is the extracted machine-readable OAuth code, never the body (which echoes the refresh token back). `snapshot(from:)` logs byte count on decode failure, not the body.
- **Nested-depth redaction (the #26 near-miss).** `flatten` walks a `JSONValue` tree and checks `redactedKeys` at *every* level, not just the root. `testOpenAIRawFieldRedaction` asserts on a `profile.organization.email` / `name` buried two levels down and confirms `profile.organization.email == "[redacted]"`. Verified nothing reaches `usage_history.raw_data` / `probe_snapshots` unredacted: `email`, `account_id`, `user_id` all redact, while `rate_limit.primary_window.limit_window_seconds` archives verbatim.
- **Explicit scratch paths (the #16 class).** Both new tests build paths from `NSTemporaryDirectory() + UUID()` and clean up in `defer` — no `$HOME` override anywhere. `RankingExporter.exportNow(dbPath:outputPath:)` and `OAuthPoller(dbPath:)` exist precisely so a test never touches the live store, and `CodexCLI`'s `--db` writes `ranking.json` beside the scratch DB rather than clobbering the real one. The `homeDirectoryForCurrentUser`-ignores-`HOME` hazard is called out in comments at all three sites.
- **No credential material in the diff.** Scanned for JWT/`sk-ant-`/`rt.`/bearer/email patterns; the only hit is the obvious dummy `"sk-ant-oat01-x"` in a self-test, and the wire fixture uses `acct-fixture` / `fixture@example.com` throughout. The synthetic JWTs are constructed inline from `{"exp":...}` literals.

One note rather than a finding: `credential.label` (which for an OpenAI row resolves to the account email) appears in the new `refreshIfNeeded` log lines. That is identical to the pre-existing Anthropic logging convention in `debug.log`, not a new leak class — flagging only so it stays a conscious choice.

### 4-5. Concurrency, Linux, selftest

The `CodexCLI.swift` `let storePath = dbPath` freeze is the right shape for the strict-concurrency error, and its comment records that CI's toolchain is the authoritative gate. I treated my local build as weak evidence per that note — CI is green on macos-14, which is what matters. Linux headless builds and `selftest` passes there too (99/99, both platforms). `OpenAIAPI.swift`/`CodexCLI.swift` are free of AppKit/SwiftUI/Combine/os.Logger, and the deliberate move from `CFBooleanGetTypeID` to a `JSONDecoder`-decoded `JSONValue` is a real portability fix, not a workaround.

### Non-blocking observations

- `RateLimit.statusString` reports `"allowed"` when both `allowed` and `limit_reached` are nil. Defensible (a present-but-empty `rate_limit` object is not evidence of rejection), but if OpenAI ever ships a shape where those go missing during an outage, the account will read as available. Worth a comment if it ever bites.
- `flatten`'s 200-field cap truncates in dictionary-iteration order, so *which* fields survive on an oversized response is nondeterministic. Only affects the moving-target archive, never a derived figure.
- The README correctly warns consumers that `email` is not unique across providers now — that is a genuine behavior change for `loom-daemon`'s join, and documenting it rather than bumping `schema` is the right trade given the field is purely additive.
- Agreed on punting `additional_rate_limits[]` charting; the data reaches `RateLimitSnapshot.named` and `probe_snapshots` today, so nothing here forecloses it.

Documentation is unusually honest about the boundary between verified and assumed — the spike doc's "Still unverified" section was narrowed to exactly what the probe actually established, rather than being quietly deleted. Approving.

PR #33, opened 2026-07-30
feat: persist and chart OpenAI per-model sub-limits (named_limits)

### LINKED ISSUE #32: Chart OpenAI per-model sub-limits (additional_rate_limits[]) in the per-account history window
## Original Issue

Follow-up to #29 (epic #25 phase 3), which listed this as optional-if-it-fits and deliberately punted it.

## What already works

The OpenAI usage response's `additional_rate_limits[]` is captured end to end today:

- `OpenAIUsageResponse.rateLimitSnapshot` files each entry into `RateLimitSnapshot.named`, keyed by `limit_name` (falling back to `metered_feature`).
- Every field is archived verbatim (PII-redacted) into `probe_snapshots.headers`, e.g. `additional_rate_limits.0.limit_name = GPT-5.3-Codex-Spark`, `...rate_limit.primary_window.used_percent = 0`.

Verified live on a ChatGPT Pro account: one entry, `GPT-5.3-Codex-Spark` / `codex_bengalfox`, with the same nested `rate_limit` shape as the account-level window.

## What's missing

Nothing surfaces it. `UsageChartView` plots `usage_history` columns, which are fixed (`session_percent` / `weekly_all_percent`) and have no room for a variable-length, provider-defined set of named sub-limits. Anthropic's analogous Fable tracking gets away with a single hard-coded `fablePercent` column read out of `probe_snapshots` at load time; that trick doesn't generalize to "N named limits whose names the provider chooses".

## Why it's worth doing

`additional_rate_limits[]` is better structured than anything Anthropic exposes — genuinely per-model, with its own window and reset, rather than an account-scoped header observed while exercising a model. It is the natural source for "which model is actually eating this account's weekly quota".

## Sketch

- A `named_limits` table (`account_id`, `timestamp`, `limit_name`, `used_percent`, `window_seconds`, `reset_at`) written alongside each `usage_history` row, rather than more fixed columns.
- A series-per-name overlay in `UsageChartView`, hidden entirely when an account has no named limits (every Anthropic account, today).
- Optionally extend `ranking.json`'s existing optional `models` object to carry them, which would keep the schema shape consumers already know.

## Notes

- Naming is provider-controlled and will churn (`GPT-5.3-Codex-Spark` is a point-in-time name) — the storage shape must not assume a fixed set.
- Same security rules as #29: no token or PII in anything charted, logged, or exported.

---

### Problem Statement (confirmed against current `main`)

Verified directly against code on `origin/main` (34e216c):

- `OpenAIUsageResponse.rateLimitSnapshot` (`OpenAIAPI.swift:140-154`) already folds every `additional_rate_limits[]` entry into `RateLimitSnapshot.named[String: RateLimitWindow]`, keyed by `AdditionalRateLimit.key` (`limit_name`, falling back to `metered_feature`).
- `RateLimitSnapshot.named` is real, documented, and already exercised in `SelfTest.swift:245` (`windows.named["GPT-5.3-Codex-Spark"]?.usedPercent`) — so the *decode* path is solid.
- But `OAuthPoller.writeSnapshotToDB` (`OAuthPoller.swift:932-944`) never reads `snapshot.rateLimit.named` at all — it only pulls `.session` / `.weekly` into the fixed `usage_history` columns (`session_percent`, `weekly_all_percent`). The named dict is dropped on the floor at the DB-write boundary.
- The only thing that survives is the flattened *verbatim* JSON archived into `probe_snapshots.headers` (e.g. `additional_rate_limits.0.limit_name`, `additional_rate_limits.0.rate_limit.primary_window.used_percent`) — a point-in-time text blob, not a queryable time series. That's what the issue means by "captured end to end" — captured for audit, not for charting.
- `UsageChartView.swift` reads history exclusively via `UsageStore.loadFullHistory()` (`UsageStore.swift:660-698`), which selects the fixed `session_percent` / `weekly_all_percent` columns into `FullUsageDataPoint` (`UsageStore.swift:171-176`). There is no code path from `probe_snapshots` (or anywhere) into the chart.

So the fix has two independent gaps to close, not one:
1. **Persistence gap** — `named` limits are computed every poll but never written anywhere queryable over time.
2. **Presentation gap** — even if they were persisted, `UsageChartView` has no series shape for a variable, provider-chosen set of named limits.

### Acceptance Criteria

- [ ] A new `named_limits` table stores one row per `(account_id, timestamp, limit_name)` with `used_percent`, `window_seconds` (nullable), and `reset_at` (nullable ISO 8601 string) — mirroring the nullable-window convention `usage_history` already uses for OpenAI (absent means *unknown*, never coerced to 0).
- [ ] `OAuthPoller.writeSnapshotToDB` writes one `named_limits` row per entry in `snapshot.rateLimit.named` on every poll, alongside the existing `usage_history` insert (same timestamp, same transaction/connection).
- [ ] Anthropic accounts continue to produce zero `named_limits` rows in normal operation (the Fable probe's `named["fable"]` entry is synthesized read-side in `UsageRecord.rateLimit`, not written by `OAuthPoller` today — confirm this stays true, or explicitly note if it now also gets a row; either is acceptable but must be a deliberate choice, not an accident).
- [ ] `UsageStore` exposes a read method (e.g. `loadNamedLimitHistory(for accountId:daysBack:)`) returning per-name time series, analogous to `loadFullHistory`.
- [ ] `UsageChartView` renders one overlay series per distinct `limit_name` seen in the account's window, **only when at least one exists** — an account with zero `named_limits` rows (every Anthropic account today, and OpenAI accounts before this ships) renders identically to today, with no empty legend or placeholder series.
- [ ] Named-limit series use provider-chosen names verbatim as the series label (e.g. "GPT-5.3-Codex-Spark") — no attempt to map them to a fixed enum, since naming is provider-controlled and will churn (per the original issue's note).
- [ ] No token, refresh token, or PII is written to `named_limits` or logged while implementing this — same rule already enforced for `probe_snapshots` (see `OpenAIUsageResponse.redactedKeys`).
- [ ] `swift build && .build/debug/ClaudeMonitor selftest` continues to pass, extended with an assertion that a decoded `additional_rate_limits[]` fixture round-trips into `named_limits` rows with correct `limit_name` / `used_percent` (portable-core test — no network, no UI).
- [ ] Linux headless build (`swift build` under `swift:6.1` + `libsqlite3-dev`) still succeeds — the persistence half of this change (schema + `OAuthPoller`) must stay in portable core; only the `UsageChartView` overlay is macOS-only (`#if os(macOS)`).

### Implementation Guidance

**Schema** (`UsageStore.swift` → `applySchema`, alongside the existing `CREATE TABLE IF NOT EXISTS` block):

```sql
CREATE TABLE IF NOT EXISTS named_limits (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    account_id TEXT NOT NULL,
    timestamp TEXT NOT NULL,
    limit_name TEXT NOT NULL,
    used_percent REAL,
    window_seconds REAL,
    reset_at TEXT,
    FOREIGN KEY (account_id) REFERENCES accounts(id)
);
CREATE INDEX IF NOT EXISTS idx_named_limits_account_time
    ON named_limits(account_id, limit_name, timestamp DESC);
```

This follows the existing idempotent-migration pattern (`applySchema` runs on every launch; `CREATE TABLE IF NOT EXISTS` heals in place, matching `usage_history` / `probe_snapshots` immediately above it in the same function).

**Writer** (`OAuthPoller.swift`): `writeSnapshotToDB` (line ~932) currently only forwards `.session` / `.weekly` to `writeUsageToDB`. Add a sibling write — after (or alongside) the `usage_history` INSERT in `writeUsageToDB` (line ~1000), loop `snapshot.rateLimit.named` and INSERT one `named_limits` row per entry, using the *same* `now` timestamp so a chart can join across tables on `(account_id, timestamp)` if ever needed. Use `RateLimitWindow.durationSeconds` / `.resetAtISO` for the two nullable columns.

**Reader** (`UsageStore.swift`): add `loadNamedLimitHistory(for:daysBack:)` mirroring `loadFullHistory`'s shape — `SELECT limit_name, timestamp, used_percent FROM named_limits WHERE account_id = ? AND timestamp >= ? ORDER BY limit_name, timestamp ASC`, grouped client-side into `[String: [FullUsageDataPoint]]` or a small dedicated struct (e.g. `NamedLimitDataPoint`).

**Chart** (`UsageChartView.swift`, macOS-only, `#if os(macOS)`): add an overlay series per distinct `limit_name`, drawn only when `loadNamedLimitHistory` returns a non-empty result for the account — this is the "hidden entirely when an account has no named limits" requirement. Reuse the existing per-series color/legend machinery already used for session/weekly traces rather than inventing a new rendering path. Because names are provider-controlled, do not hardcode a palette keyed by name — assign colors positionally/deterministically (e.g. hash of the name, or index into a fixed palette by sorted-name order) so a renamed model doesn't silently reuse another model's prior color in a way that reads as continuity.

**Out of scope for this issue** (explicitly deferred, per the original issue's own "Optionally" framing): extending `ranking.json`'s `models` object to carry named limits (`RankingExporter.swift`). That's a separate, smaller follow-up once the storage shape here is settled — file it separately if still wanted after this lands, rather than bundling an export-schema change into the same PR as a new table + writer + chart overlay.

### Affected Files

- `menubar-app/ClaudeMonitor/Sources/UsageStore.swift` - add `named_limits` table to `applySchema`; add `loadNamedLimitHistory(for:daysBack:)` (or equivalent) read method
- `menubar-app/ClaudeMonitor/Sources/OAuthPoller.swift` - `writeSnapshotToDB` / `writeUsageToDB` (~lines 932-1017): write one `named_limits` row per `snapshot.rateLimit.named` entry per poll
- `menubar-app/ClaudeMonitor/Sources/UsageChartView.swift` - add a per-`limit_name` overlay series, macOS-only, hidden when the account has no named limits
- `menubar-app/ClaudeMonitor/Sources/SelfTest.swift` - extend portable-core coverage: decode a fixture with `additional_rate_limits[]`, verify it round-trips into `named_limits` rows
- `menubar-app/ClaudeMonitor/Sources/RateLimitWindow.swift` - reference only; no change expected unless the writer needs a helper this file doesn't already expose

### Test Plan

- [ ] Automated: `swift build && .build/debug/ClaudeMonitor selftest` passes, including the new named-limits round-trip assertion (decode fixture → write → read back → compare `limit_name` / `used_percent`).
- [ ] Automated: Linux build verified via `swift:6.1` Docker image (`apt-get install libsqlite3-dev && swift build`) — the schema + writer changes must stay in portable core.
- [ ] Manual verification (macOS, ChatGPT Pro/Codex account with an active `additional_rate_limits[]` entry): poll at least twice, open the account's usage-history chart window, confirm a named-limit overlay series appears and updates.
- [ ] Manual verification (Anthropic account, or any account with zero named limits): confirm the chart renders exactly as it did before this change — no empty series, no legend entry, no layout shift.
- [ ] Edge case: a named limit whose `limit_name` changes between polls (provider renames a model) should appear as two distinct series rather than a single series with a discontinuous name — confirm this is the natural behavior of keying by `limit_name` rather than by a stable synthetic id.
- [ ] Edge case: confirm no token, refresh token, or raw identity field appears in the new `named_limits` table or in any log line touched by this change.

rjwalters:
## Summary

- Adds a `named_limits` table (`account_id`, `timestamp`, `limit_name`, `used_percent`, `window_seconds`, `reset_at`) to `applySchema`, alongside its own index, so `RateLimitSnapshot.named` (OpenAI's `additional_rate_limits[]`) survives as a queryable time series rather than only a verbatim `probe_snapshots.headers` JSON blob.
- `OAuthPoller.writeUsageToDB` now writes one `named_limits` row per entry in the snapshot's `named` map, stamped with the same timestamp as the sibling `usage_history` row. Both `writeSnapshotToDB` (OpenAI) and `writePingToDB` (Anthropic) pass their respective `rateLimit.named` map through — Anthropic's ping response never populates `named`, so this remains a true no-op for those accounts.
- `UsageStore.loadNamedLimitHistory(for:daysBack:)` reads the table back, grouped by `limit_name`, ascending by timestamp.
- `UsageChartView` renders one overlay series per distinct `limit_name`, using the provider's name verbatim as the label, entirely hidden (no legend, no empty series) when an account has zero named limits.
- `SelfTest.swift` adds `testNamedLimitsRoundTrip`: decodes the existing `additional_rate_limits[]` fixture, writes it via `UsageStore.insertNamedLimits`, reads it back, and confirms `limit_name`/`used_percent` round-trip; also asserts an empty `named` map writes zero rows and an account with no rows reads back empty.

## Changes

- `menubar-app/ClaudeMonitor/Sources/UsageStore.swift` — `named_limits` table + index in `applySchema`; `NamedLimitDataPoint` struct; `UsageStore.insertNamedLimits` (static writer, shared by `OAuthPoller` and the self-test); `loadNamedLimitHistory(for:daysBack:)` reader.
- `menubar-app/ClaudeMonitor/Sources/OAuthPoller.swift` — `writeUsageToDB` gains a `namedLimits` parameter (default `[:]`); `writeSnapshotToDB` / `writePingToDB` pass `rateLimit.named` through.
- `menubar-app/ClaudeMonitor/Sources/UsageChartView.swift` — `NamedLimitTrace` type; a dedicated `namedLimitColors` palette assigned positionally by sorted `limit_name`; overlay series + legend in the percent chart, gated on `!namedLimitTraces.isEmpty`; `ChartWindowController.showChart` loads and builds the traces.
- `menubar-app/ClaudeMonitor/Sources/SelfTest.swift` — `testNamedLimitsRoundTrip`.

`RateLimitWindow.swift` was reviewed per the issue's Affected Files list but needed no change — its existing `durationSeconds` / `resetAtISO` accessors already cover the writer's needs.

- [x] `named_limits` table with nullable `window_seconds` / `reset_at`, mirroring the existing nullable-window convention.
- [x] `OAuthPoller.writeSnapshotToDB` (and `writePingToDB`) write one row per `named` entry alongside the `usage_history` insert, same timestamp.
- [x] Anthropic accounts write zero `named_limits` rows — confirmed both by code inspection (ping's `rateLimit.named` is always `[:]`) and by an explicit selftest assertion (`insertNamedLimits` with an empty map writes zero rows).
- [x] `UsageStore.loadNamedLimitHistory(for:daysBack:)` added, analogous to `loadFullHistory`.
- [x] `UsageChartView` renders one overlay per distinct `limit_name`, only when at least one exists; verified by gating both the chart `ForEach` and the legend `HStack` on `!namedLimitTraces.isEmpty` / `!filteredNamedLimitTraces.isEmpty`.
- [x] Series labels use the provider name verbatim — no enum mapping.
- [x] No token/PII touched — the writer only consumes the already-decoded, already-redacted `RateLimitWindow` values.
- [x] `swift build && .build/debug/ClaudeMonitor selftest` passes (105 checks, up from 104), including the new round-trip assertion.
- [x] Linux headless build verified via `swift:6.1` Docker image (`apt-get install libsqlite3-dev && swift build`) — same 105 checks pass. Schema + writer changes are in portable core; only `UsageChartView` is `#if os(macOS)`.
- [x] `ranking.json` extension explicitly left out of scope per the issue's own deferral.

## Test Plan

- [x] `swift build && .build/debug/ClaudeMonitor selftest` — 105/105 checks pass.
- [x] Linux build + selftest via `swift:6.1` Docker image — 105/105 checks pass.
- [ ] Manual verification (ChatGPT Pro/Codex account with an active `additional_rate_limits[]` entry): poll at least twice, open the chart window, confirm the overlay appears and updates. (Not performed in this PR — no live OpenAI/Codex credential available in this environment; covered by the automated fixture round-trip instead.)
- [x] Edge case: an account with zero named limits renders identically to before — verified via the empty-history selftest assertion and the `!namedLimitTraces.isEmpty` gates in the view.
- [x] Edge case: a renamed limit naturally appears as a second series (keyed by `limit_name`, not a synthetic id) — this falls out of the dictionary-keyed storage/read shape, no special-casing needed.
- [x] No token/PII in the new table or any touched log line — `insertNamedLimits` only ever receives `RateLimitWindow` values, which carry no identity fields.

Closes #32

### DIFF
--- a/menubar-app/ClaudeMonitor/Sources/OAuthPoller.swift
+++ b/menubar-app/ClaudeMonitor/Sources/OAuthPoller.swift
@@ -919,7 +919,8 @@ class OAuthPoller: ObservableObject {
             weeklyReset: windows.weekly?.resetAtISO,
             rawFields: ping.rawHeaders,
             probeModel: "haiku",
-            httpStatus: ping.httpStatus
+            httpStatus: ping.httpStatus,
+            namedLimits: windows.named
         )
     }
 
@@ -939,12 +940,18 @@ class OAuthPoller: ObservableObject {
             weeklyReset: windows.weekly?.resetAtISO,
             rawFields: snapshot.rawFields,
             probeModel: "\(snapshot.provider.rawValue)-usage",
-            httpStatus: snapshot.httpStatus
+            httpStatus: snapshot.httpStatus,
+            namedLimits: windows.named
         )
     }
 
     /// The single write path for both providers: one `usage_history` row plus a
     /// verbatim `probe_snapshots` archive entry.
+    ///
+    /// `namedLimits` additionally writes one `named_limits` row per entry
+    /// (OpenAI's `additional_rate_limits[]`), stamped with the same timestamp
+    /// as the `usage_history` row. Empty for Anthropic pings today, so those
+    /// accounts continue to produce zero `named_limits` rows.
     private func writeUsageToDB(
         accountId: String,
         sessionPercent: Double?,
@@ -953,7 +960,8 @@ class OAuthPoller: ObservableObject {
         weeklyReset: String?,
         rawFields: [String: String],
         probeModel: String,
-        httpStatus: Int
+        httpStatus: Int,
+        namedLimits: [String: RateLimitWindow] = [:]
     ) {
         guard FileManager.default.fileExists(atPath: dbPath) else { return }
 
@@ -1004,6 +1012,8 @@ class OAuthPoller: ObservableObject {
                 ) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, 0)
             """, accountId, now, primaryPercent, sessionPercent,
                weeklyPercent, 0.0, sessionReset, weeklyReset, rawData)
+
+            UsageStore.insertNamedLimits(db, accountId: accountId, timestamp: now, named: namedLimits)
 
             try db.run("UPDATE accounts SET last_updated = ? WHERE id = ?", now, accountId)
 
--- a/menubar-app/ClaudeMonitor/Sources/SelfTest.swift
+++ b/menubar-app/ClaudeMonitor/Sources/SelfTest.swift
@@ -49,6 +49,7 @@ enum SelfTest {
         testCodexAuthParsing()
         testSchemaMigrationFromPreMigrationDatabase()
         testRankingExportCarriesProvider()
+        testNamedLimitsRoundTrip()
 
         if let idx = arguments.firstIndex(of: "--db"), idx + 1 < arguments.count {
             testMigrationOfExistingDatabase(at: arguments[idx + 1])
@@ -430,6 +431,60 @@ enum SelfTest {
         } catch {
             checks += 1
             failures.append("Codex auth parsing test threw: \(error)")
+        }
+    }
+
+    // MARK: - Named limits (per-model sub-limits, #32)
+
+    /// Decodes a fixture carrying `additional_rate_limits[]`, writes the
+    /// resulting `named` map into a throwaway database via
+    /// `UsageStore.insertNamedLimits` (the same helper `OAuthPoller` calls on
+    /// every poll), then reads it back via `loadNamedLimitHistory` and
+    /// confirms `limit_name` / `used_percent` survive the round trip.
+    private static func testNamedLimitsRoundTrip() {
+        let dir = URL(fileURLWithPath: NSTemporaryDirectory())
+            .appendingPathComponent("claude-monitor-selftest-\(UUID().uuidString)")
+        defer { try? FileManager.default.removeItem(at: dir) }
+        do {
+            try FileManager.default.createDirectory(at: dir, withIntermediateDirectories: true)
+            let dbPath = dir.appendingPathComponent("usage.db").path
+
+            let store = UsageStore(dbPath: dbPath)
+            store.ensureDatabase()
+
+            let snapshot = try OpenAIAPIClient.snapshot(
+                from: Data(openAIUsageFixture.utf8), httpStatus: 200
+            )
+            let named = snapshot.rateLimit.named
+            expect(!named.isEmpty, "fixture must decode at least one named sub-limit")
+
+            let db = try openDatabase(dbPath)
+            let now = ISO8601DateFormatter().string(from: Date())
+            UsageStore.insertNamedLimits(db, accountId: "acct-fixture", timestamp: now, named: named)
+
+            let history = store.loadNamedLimitHistory(for: "acct-fixture")
+            expectEqual(history.count, 1, "one series per distinct limit_name")
+            let series = history["GPT-5.3-Codex-Spark"]
+            expect(series != nil, "the fixture's limit_name is preserved verbatim as the series key")
+            expectEqual(series?.first?.usedPercent, 62,
+                        "used_percent round-trips through named_limits")
+
+            // An account with no named limits at all must read back empty —
+            // this is what lets the chart overlay stay hidden for every
+            // Anthropic account.
+            let emptyHistory = store.loadNamedLimitHistory(for: "acct-with-no-named-limits")
+            expect(emptyHistory.isEmpty, "an account with zero named_limits rows reads back an empty dictionary")
+
+            // Anthropic's ping response never populates `named` — confirm the
+            // write path is a true no-op for it, not merely untested.
+            UsageStore.insertNamedLimits(db, accountId: "acct-anthropic", timestamp: now, named: [:])
+            let anthropicCount = try db.scalar(
+                "SELECT COUNT(*) FROM named_limits WHERE account_id = 'acct-anthropic'"
+            ) as? Int64
+            expectEqual(anthropicCount, 0, "an empty named map writes zero named_limits rows")
+        } catch {
+            checks += 1
+            failures.append("named limits round-trip test threw: \(error)")
         }
     }
 
--- a/menubar-app/ClaudeMonitor/Sources/UsageChartView.swift
+++ b/menubar-app/ClaudeMonitor/Sources/UsageChartView.swift
@@ -26,6 +26,16 @@ struct TokenTrace: Identifiable {
     let isPrimary: Bool
 }
 
+/// One named sub-limit's overlay series (e.g. an OpenAI
+/// `additional_rate_limits[]` entry). Named by the provider — the label is
+/// used verbatim, never mapped to a fixed enum, since naming will churn.
+struct NamedLimitTrace: Identifiable {
+    let id: String  // the provider's limit_name, doubling as a stable-enough series key
+    let name: String
+    let dataPoints: [NamedLimitDataPoint]
+    let color: Color
+}
+
 struct UsageChartWindow: View {
     let account: Account
     let dataPoints: [UsageDataPoint]
@@ -35,6 +45,7 @@ struct UsageChartWindow: View {
     let oauthPoller: OAuthPoller?
     let otherAccountsData: [AccountTrace]  // Data for other accounts
     let otherTokenData: [TokenTrace]  // Token data for other accounts
+    let namedLimitTraces: [NamedLimitTrace]  // Per-model sub-limits for this account (empty for Anthropic)
     @Environment(\.colorScheme) var colorScheme
     @StateObject private var updateChecker = UpdateChecker.shared
     @State private var isEditingName = false
@@ -125,6 +136,23 @@ struct UsageChartWindow: View {
                 dataPoints: trace.dataPoints.filter { $0.timestamp >= startDate && $0.timestamp <= endDate },
                 color: trace.color,
                 isPrimary: trace.isPrimary
+            )
+        }
+    }
+
+    /// Filtered named-limit traces based on the current range selection.
+    /// Empty when the account has no named limits — callers must check this
+    /// before rendering anything so the overlay stays entirely hidden.
+    var filteredNamedLimitTraces: [NamedLimitTrace] {
+        guard !namedLimitTraces.isEmpty else { return [] }
+        let startDate = dateForRangePosition(rangeStart)
+        let endDate = dateForRangePosition(rangeEnd)
+        return namedLimitTraces.map { trace in
+            NamedLimitTrace(
+                id: trace.id,
+                name: trace.name,
+                dataPoints: trace.dataPoints.filter { $0.timestamp >= startDate && $0.timestamp <= endDate },
+                color: trace.color
             )
         }
     }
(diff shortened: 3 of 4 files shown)


rjwalters:
✅ **Approved!**

Reviewed against issue #32's acceptance criteria and verified independently (not just from the PR description).

**Correctness**
- `named_limits` table (`account_id, timestamp, limit_name, used_percent, window_seconds, reset_at`) added via idempotent `CREATE TABLE IF NOT EXISTS` in `applySchema` — nullable `window_seconds`/`reset_at` correctly mirror the existing nullable-window convention.
- `UsageStore.insertNamedLimits` is a shared static writer called from both `writePingToDB` (Anthropic) and `writeSnapshotToDB` (OpenAI) via `writeUsageToDB`'s new `namedLimits` param, stamped with the same `now` timestamp as the sibling `usage_history` row.
- Confirmed `RateLimitSnapshot.named` is genuinely empty for Anthropic pings (never populated in `OAuthPoller.swift`), so the write path is a true no-op there — also explicitly asserted in `testNamedLimitsRoundTrip`.
- `loadNamedLimitHistory` mirrors `loadFullHistory`'s shape/conventions (string-comparison timestamp filtering, `parseDate` fallback for fractional/non-fractional ISO8601 — both pre-existing, reused correctly here).
- `UsageChartView`'s overlay is correctly nested inside the existing `chartMode == .percent` block, follows the same `ForEach(trace) { ForEach(points) { LineMark/PointMark } }` pattern as the pre-existing `otherAccountsData` overlay, and both the chart `ForEach` and legend `HStack` are gated on non-empty traces so an Anthropic account (or any account with zero named limits) renders identically to before.
- Colors are assigned positionally by sorted `limit_name` (not by hashing/name-based lookup), matching the issue's guidance to avoid a renamed model silently inheriting another model's prior color.

**Security / PII**
- `AdditionalRateLimit.key` (`limit_name`/`metered_feature`) is a model/feature label (e.g. `GPT-5.3-Codex-Spark`), not an identity field — confirmed by inspecting `OpenAIAPI.swift`. `insertNamedLimits` only ever receives already-decoded `RateLimitWindow` values and logs only `limitName` + a generic error description on failure — no tokens or raw identity fields touched.

**Portability**
- `UsageChartView.swift` remains entirely `#if os(macOS)`-fenced (verified: `#if os(macOS)` at line 1, matching `#endif` at EOF).
- `UsageStore.swift`, `OAuthPoller.swift`, `SelfTest.swift` (the portable-core files touched) have zero AppKit/SwiftUI/Combine/os.Logger imports — portable core stays portable.

## Test Execution

- `swift build && .build/debug/ClaudeMonitor selftest` (macOS, local) — ✅ 105/105 checks pass.
- Linux build + selftest via `swift:6.1` Docker image (fresh checkout, `libsqlite3-dev`) — ✅ 105/105 checks pass (only pre-existing, unrelated Swift 6 concurrency warnings in `UsageStore.swift`'s update-checker code, not touched by this PR).
- CI (`gh pr checks`): `build-macos` pass, `build-linux` pass. `mergeStateStatus`: CLEAN.
- Manual live-OpenAI-account verification step in the PR's test plan is correctly flagged as not performed (no live credential in this environment) — acceptable since it's covered by the automated fixture round-trip (`testNamedLimitsRoundTrip`), consistent with the issue's own edge-case criteria.

Nicely scoped — the `ranking.json` extension was correctly left out per the issue's own deferral, and the PR closes #32 with the proper `Closes #32` syntax.

PR #42, opened 2026-07-31
docs: clarify local swift build does not gate strict-concurrency errors

### LINKED ISSUE #41: Document that a clean local swift build does not gate Swift strict-concurrency errors
## Context

PR #31 (Phase 3) failed CI on `build-macos` with:

```
CodexCLI.swift:72: error: mutation of captured var 'result' in concurrently-executing code
```

while **`build-linux` passed**. During the fix, the Doctor found something worth recording as a standing rule: its own local macOS toolchain (Swift 6.3.3) also *accepted* the broken code — newer toolchains downgraded this diagnostic to a warning under region-based isolation, while CI's `macos-14` Xcode still errors on it.

So the Doctor's first "verified locally, clean build" was a **false green**, and it only caught the second latent instance (`reference to captured var 'dbPath'`) after CI failed a second time. Its workaround was `swift build -Xswiftc -swift-version -Xswiftc 6` as a strictly-stronger local proxy, validated by confirming the proxy errors on the original code.

This is a trap any future contributor (human or agent) will fall into: `swift build` succeeding locally is normally taken as sufficient evidence, and for this bug class it isn't.

- [ ] `CLAUDE.md` records that for Swift concurrency diagnostics, **CI is authoritative** and a clean local `swift build` is weak evidence, naming the toolchain-version divergence as the reason.
- [ ] It documents the stronger local proxy (`swift build -Xswiftc -swift-version -Xswiftc 6`) so the check can be run before pushing.
- [ ] Consider whether the package should simply adopt Swift 6 language mode so local and CI agree by construction — worth evaluating, but a bigger change than the doc note and should not block it.

rjwalters:
## Summary
- Adds a note to `CLAUDE.md`'s Headless/Linux section documenting that a clean local `swift build` is weak evidence for Swift strict-concurrency diagnostics (e.g. captured-var mutation in concurrent code): newer local toolchains (6.3.3+) can downgrade these to warnings under region-based isolation, while CI's pinned `macos-14` Xcode still hard-errors on them.
- Documents the stronger local proxy check to run before pushing: `swift build -Xswiftc -swift-version -Xswiftc 6`.
- Notes (without mandating) that adopting Swift 6 language mode package-wide would make local and CI agree by construction — a bigger change left for future consideration.

Documentation-only change; no code touched.

Closes #41

## Test plan
- [x] Reviewed rendered CLAUDE.md for tone/format consistency with surrounding bullet points
- N/A build/test — docs-only change

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -22,6 +22,7 @@
 - To verify a schema migration against real data, copy the live DB first — `cp ~/.claude-monitor/usage.db /tmp/ && .build/debug/ClaudeMonitor selftest --db /tmp/usage.db`. `--db` **writes** to the path it is given; never point it at `~/.claude-monitor/usage.db`.
 - Parse response headers via `extractAnthropicHeaders` (lowercased map), never `allHeaderFields` subscripts — those are case-sensitive on Linux.
 - Verify Linux builds from macOS with the `swift:6.1` Docker image (`apt-get install libsqlite3-dev`, then `swift build`).
+- **A clean local `swift build` is weak evidence for Swift strict-concurrency errors** (e.g. "mutation of captured var in concurrently-executing code") — a newer local toolchain (6.3.3+) can downgrade these to warnings under region-based isolation while CI's pinned `macos-14` Xcode still hard-errors on them. CI is authoritative for this bug class. Before pushing, run the stronger local proxy `swift build -Xswiftc -swift-version -Xswiftc 6` (forces Swift 6 language mode, surfacing the same diagnostics CI's toolchain flags). Adopting Swift 6 language mode package-wide would make local and CI agree by construction, but that's a bigger change than this note and hasn't been done yet.
 
 ## Project Structure
 


rjwalters:
LGTM — approving.

Verified against issue #41's acceptance criteria:
- (a) States CI is authoritative for this diagnostic class and names the cause: "a newer local toolchain (6.3.3+) can downgrade these to warnings under region-based isolation while CI's pinned macos-14 Xcode still hard-errors on them."
- (b) Documents the stronger local proxy check verbatim: `swift build -Xswiftc -swift-version -Xswiftc 6`.
- (c) Correctly does *not* mandate package-wide Swift 6 language mode — phrased as an optional future consideration ("a bigger change than this note and hasn't been done yet"), matching the issue's explicit scope-out.

Other checks:
- Diff is CLAUDE.md only (+1/-0), no unrelated files touched.
- New bullet is placed sensibly in the Headless/Linux section, right after the existing Linux-build-verification bullet — topically adjacent.
- Style/tone matches surrounding bullets (bold lead-in like the "Important:" bullet, backtick-quoted commands, terse).
- CI green: `build-macos` and `build-linux` both pass; `mergeStateStatus` is CLEAN.

No changes requested.

PR #43, opened 2026-07-31
fix: tiebreak account ordering on natural display name, not account id

### LINKED ISSUE #40: sortedAccountsForPopover tie-breaks on account UUID, giving arbitrary order
## Context

`UsageStore.sortedAccountsForPopover` ranks accounts by availability, then breaks ties on reset time. When those are also equal — common for freshly-added or idle accounts that all read 0% — ordering falls through to insertion/UUID order, which is arbitrary and unstable to a human reading the list.

This is the default popover ordering **and** what the headless Linux daemon prints, so the arbitrariness shows up in two places.

Related: `UsagePopoverView.compareRows` has the same `a.0.id < b.0.id` fallback for its own tie cases.

PR #34 added `NaturalSort.compare` (hybrid lexical/numeric, so `agent-10` follows `agent-9`) and wired it into the Account **column** sort only. Reusing it as the final tiebreak everywhere would make equal-usage accounts order predictably by name.

Deliberately left out of #34: `UsageStore.swift` was being edited concurrently by the in-flight #32 sweep, and a same-file collision wasn't worth the rebase. That sweep has since merged, so the file is free.

- [ ] `sortedAccountsForPopover` falls back to `NaturalSort.compare` on display name instead of account id when usage and reset times tie.
- [ ] `compareRows`' tie fallbacks do the same, so the table and the default order agree.
- [ ] `NaturalSort` stays in the portable core so the headless Linux target still builds (it already is, and is Foundation-only).
- [ ] A selftest pins that two accounts with identical usage order by natural name.

## Affected Files

- `menubar-app/ClaudeMonitor/Sources/UsageStore.swift`
- `menubar-app/ClaudeMonitor/Sources/UsagePopoverView.swift`
- `menubar-app/ClaudeMonitor/Sources/SelfTest.swift`

rjwalters:
## Summary

- `UsageStore.sortedAccountsForPopover` fell through to arbitrary insertion/UUID order (`a.account.id < b.account.id`) when usage percent and reset time both tied — common for freshly-added or idle accounts that all read 0%.
- `UsagePopoverView.compareRows` had the same `a.0.id < b.0.id` fallback in three places (the `.account` column's own tie, the `(nil, nil)` case, and the `x == y` case for every other column).
- PR #34 added `NaturalSort.compare` (hybrid lexical/numeric ordering, so `agent-10` follows `agent-9`) but only wired it into the Account column's primary sort.

## Changes

- `UsageStore.swift`: `sortedAccountsForPopover` now falls back to `NaturalSort.compare` on display name after usage and reset time tie, before falling to account id as the final total-order tiebreak.
- `UsagePopoverView.swift`: extracted a shared `stableTiebreak(_:_:)` helper (natural name, then id) and used it at all three tie-fallback sites in `compareRows`, so the table's default order and the popover's default order agree.
- `SelfTest.swift`: added `testSortedAccountsForPopoverTieBreak`, which seeds two accounts with identical usage/reset but ids that *disagree* with natural name order, and asserts `sortedAccountsForPopover` still orders them `agent-9` before `agent-10`.

- [x] `sortedAccountsForPopover` falls back to `NaturalSort.compare` on display name instead of account id when usage and reset times tie — verified by the new selftest.
- [x] `compareRows`'s tie fallbacks do the same, so the table and the default order agree — all three `a.0.id < b.0.id` sites now route through `stableTiebreak`.
- [x] `NaturalSort` stays in the portable core — unchanged, still Foundation-only; the headless Linux build was re-verified end to end.
- [x] A selftest pins that two accounts with identical usage order by natural name — `testSortedAccountsForPopoverTieBreak` in `SelfTest.swift`.

## Test Plan

- `swift build` — clean build, no new warnings.
- `.build/debug/ClaudeMonitor selftest` — 116/116 checks pass (macOS).
- Verified the same build + selftest inside the `swift:6.1` Docker image (`apt-get install libsqlite3-dev`) — 116/116 checks pass on Linux too, confirming the headless daemon target is unaffected.

Closes #40

### DIFF
--- a/menubar-app/ClaudeMonitor/Sources/SelfTest.swift
+++ b/menubar-app/ClaudeMonitor/Sources/SelfTest.swift
@@ -40,6 +40,7 @@ enum SelfTest {
         }
 
         testNaturalSort()
+        testSortedAccountsForPopoverTieBreak()
         testWindowKindDerivation()
         testSnapshotFromPositionalWindows()
         testMissingSessionWindow()
@@ -95,6 +96,40 @@ enum SelfTest {
         let result = NaturalSort.selfCheck()
         checks += result.checks
         failures.append(contentsOf: result.failures.map { "naturalSort: \($0)" })
+    }
+
+    /// Two accounts with identical usage and reset must not fall through to
+    /// arbitrary insertion/UUID order (#40): the account id here is chosen to
+    /// *disagree* with natural name order, so a fix that still tiebreaks on
+    /// id would sort these the wrong way.
+    private static func testSortedAccountsForPopoverTieBreak() {
+        let store = UsageStore(dbPath: ":memory:selftest-tiebreak")
+
+        let agentNine = Account(
+            id: "zzz-later-id", accountName: "agent-9", email: nil, plan: "Max",
+            lastUpdated: nil, latestPercent: nil
+        )
+        let agentTen = Account(
+            id: "aaa-earlier-id", accountName: "agent-10", email: nil, plan: "Max",
+            lastUpdated: nil, latestPercent: nil
+        )
+        store.accounts = [agentTen, agentNine]
+
+        func identicalUsage(_ accountId: String, _ recordId: Int64) -> UsageRecord {
+            UsageRecord(
+                id: recordId, accountId: accountId, timestamp: Date(),
+                primaryPercent: 50, sessionPercent: 50, weeklyAllPercent: 20,
+                weeklySONnetPercent: nil, sessionReset: nil, weeklyReset: nil
+            )
+        }
+        store.latestUsage = [
+            agentNine.id: identicalUsage(agentNine.id, 1),
+            agentTen.id: identicalUsage(agentTen.id, 2),
+        ]
+
+        let ordered = store.sortedAccountsForPopover.map { $0.displayName }
+        expectEqual(ordered, ["agent-9", "agent-10"],
+                    "equal usage/reset falls back to natural name order, not account id")
     }
 
     // MARK: - Window model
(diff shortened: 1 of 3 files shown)


rjwalters:
Approved.

**Correctness against #40's acceptance criteria** — all four verified directly:
- `sortedAccountsForPopover` (`UsageStore.swift`) now falls back to `NaturalSort.compare` on display name after the usage/reset-time tie, before finally falling to `account.id` as a last-resort total-order tiebreak (correct — `NaturalSort` only guarantees `orderedSame` for equal-content strings, so keeping the id fallback is the right belt-and-suspenders move, not dead code).
- `UsagePopoverView.compareRows`'s three `a.0.id < b.0.id` tie sites (the `.account` column's own tie, the `(nil, nil)` case, and the `x == y` case) are now unified behind a single `stableTiebreak(_:_:)` helper — good de-duplication versus repeating the same three-line comparator three times, and it keeps the table's default order consistent with the popover's default order as the issue asked.
- `NaturalSort` itself is untouched, stays Foundation-only in the portable core — no `#if os(macOS)` violations introduced.
- New `testSortedAccountsForPopoverTieBreak` in `SelfTest.swift` seeds `agent-9`/`agent-10` with ids that *disagree* with natural-name order (`zzz-later-id` vs `aaa-earlier-id`) and identical usage/reset, which is exactly the right adversarial construction — a regression that reintroduces id-based tiebreaking would fail this test, not just pass it by accident.

**Verification performed:**
- `swift build` — clean, no warnings.
- `.build/debug/ClaudeMonitor selftest` — 116/116 checks pass locally, matching the PR description.
- `gh pr checks 43` — both `build-macos` and `build-linux` CI jobs pass.
- `mergeStateStatus` — `CLEAN`.

**Code quality:** the `stableTiebreak` extraction is a nice touch — it turns three near-duplicate id-fallback sites into one documented helper, addressing exactly the drift the issue called out (PR #34 only wired natural sort into the Account column). Comments are clear about *why* the tie is common (freshly-added/idle 0% accounts) rather than just *what* the code does.

No blocking issues found.

PR #44, opened 2026-07-31
fix: fall back to claude --version for User-Agent detection

### LINKED ISSUE #39: build-macos-app.sh User-Agent detection misses a Homebrew-installed claude
## Context

`scripts/build-macos-app.sh` auto-detects the installed Claude Code version to patch `claudeCodeUserAgent` in `AnthropicAPI.swift`:

```bash
CLAUDE_CODE_VERSION=$(npm list -g @anthropic-ai/claude-code --depth=0 2>/dev/null | grep claude-code | sed 's/.*@//')
```

On this machine `claude` is installed via **Homebrew**, not npm, so the npm query returns nothing. The script then takes its `else` branch — `Warning: could not detect claude-code version, using existing value` — and silently ships whatever string is already in the source.

**This already bit us.** At the v1.17.0 release the User-Agent was still `claude-code/2.0.37` while the installed CLI was `2.1.220`; it was corrected by hand in the release commit. Left alone it will drift again on every subsequent release, and the warning is easy to miss in build output.

## Repro

```
$ npm list -g @anthropic-ai/claude-code --depth=0   # → empty
$ claude --version                                   # → 2.1.220 (Claude Code)
$ grep claudeCodeUserAgent menubar-app/ClaudeMonitor/Sources/AnthropicAPI.swift
```

- [ ] Detection falls back to `claude --version` (parsing the leading semver) when the npm query yields nothing.
- [ ] When no version can be detected by *any* means, the warning is loud enough not to be lost — or the build fails, since shipping a stale User-Agent is a silent correctness problem, not a cosmetic one.
- [ ] Optionally: the release checklist (`.claude/commands/release.md`) notes verifying the patched User-Agent, since it's a release-time correctness item.

rjwalters:
## Summary

`scripts/build-macos-app.sh` auto-detects the installed `claude-code` version to patch `claudeCodeUserAgent` in `AnthropicAPI.swift` before compiling. The detection only queried `npm list -g @anthropic-ai/claude-code`, which returns nothing on machines where `claude` is installed via Homebrew — the script then silently kept whatever User-Agent string was already committed, with only an easy-to-miss warning. This already caused a real bug at the v1.17.0 release (stale `claude-code/2.0.37` shipped while the installed CLI was `2.1.220`).

## Changes

- `scripts/build-macos-app.sh`: when the npm query yields nothing, fall back to parsing the leading semver out of `claude --version` (e.g. `2.1.220 (Claude Code)` -> `2.1.220`).
- When neither the npm query nor `claude --version` can detect a version, the build now fails outright (`exit 1`) instead of silently proceeding with a stale value — shipping a stale User-Agent is a correctness bug, not cosmetic. `ALLOW_STALE_USER_AGENT=1` opts into building anyway for the rare case that's deliberate.
- `.claude/commands/release.md`: release checklist now has an explicit step to verify the patched `claudeCodeUserAgent` value against the `claude-code` version intended for shipping, and notes the fallback + hard-failure behavior.

| Detection falls back to `claude --version` when npm query yields nothing | ✅ | On this machine `npm list -g @anthropic-ai/claude-code` returns empty (Homebrew install); running `./scripts/build-macos-app.sh` printed `Detected claude-code version: 2.1.220`, matching `claude --version` output, confirming the fallback path fired |
| Failure to detect by any means is loud / build-failing, not a silent warning | ✅ | Manually simulated the empty-detection branch in isolation — script prints an `ERROR:` message on stderr and `exit 1`s unless `ALLOW_STALE_USER_AGENT=1` is set |
| (Optional) release checklist notes verifying the patched User-Agent | ✅ | Added an explicit verification step + note in `.claude/commands/release.md` Phase 7 |

## Test Plan

- `bash -n scripts/build-macos-app.sh` — syntax check passes
- Ran `./scripts/build-macos-app.sh` end-to-end on a machine with a Homebrew-only `claude` install (no npm global package): detected `2.1.220` via the `claude --version` fallback, patched (already-correct) User-Agent, built `build/ClaudeMonitor.app` successfully
- `cd menubar-app/ClaudeMonitor && swift build && .build/debug/ClaudeMonitor selftest` — build succeeds, `selftest: 115 check(s) passed`
- Manually verified the hard-failure branch (`exit 1` on stderr `ERROR:` message) in an isolated bash snippet mirroring the script's new detection logic

Closes #39

### DIFF
--- a/.claude/commands/release.md
+++ b/.claude/commands/release.md
@@ -159,11 +159,16 @@ The build is local — there is no release workflow that produces the `.zip`. Th
    ```bash
    ./scripts/build-macos-app.sh
    ```
-   This produces `build/ClaudeMonitor.app` and `build/ClaudeMonitor.zip`. The script auto-detects the installed `claude-code` version (via npm) and patches the User-Agent in `AnthropicAPI.swift` before compiling — confirm with the user that the local `claude-code` is on the version they want to ship with.
+   This produces `build/ClaudeMonitor.app` and `build/ClaudeMonitor.zip`. The script auto-detects the installed `claude-code` version (via `npm list -g`, falling back to `claude --version` for Homebrew/other installs) and patches the User-Agent in `AnthropicAPI.swift` before compiling — confirm with the user that the local `claude-code` is on the version they want to ship with. If neither detection method finds a version, the script now fails outright rather than silently shipping a stale User-Agent (set `ALLOW_STALE_USER_AGENT=1` to override deliberately).
 
-2. **Verify the `.zip` exists** at `build/ClaudeMonitor.zip` and is non-trivially sized.
+2. **Verify the patched User-Agent** matches the version the script reported (and the `claude-code` version you intend to ship with):
+   ```bash
+   grep claudeCodeUserAgent menubar-app/ClaudeMonitor/Sources/AnthropicAPI.swift
+   ```
 
-3. **Create the GitHub Release** using the CHANGELOG entry as the body:
+3. **Verify the `.zip` exists** at `build/ClaudeMonitor.zip` and is non-trivially sized.
+
+4. **Create the GitHub Release** using the CHANGELOG entry as the body:
    ```bash
    gh release create vX.Y.Z \
      --title "vX.Y.Z" \
@@ -172,7 +177,7 @@ The build is local — there is no release workflow that produces the `.zip`. Th
    ```
    (Manually extract the new entry from CHANGELOG.md if the awk doesn't pick up cleanly — past releases have the body matching the `### Summary` + section bullets, without the `## [X.Y.Z] - DATE` header.)
 
-4. **Verify the release page** shows the `.zip` asset and that the notes render correctly:
+5. **Verify the release page** shows the `.zip` asset and that the notes render correctly:
    ```bash
    gh release view vX.Y.Z --json name,assets,url --jq '.'
    ```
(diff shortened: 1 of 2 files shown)


rjwalters:
Approved.

Verified this actually fixes #39, not just cosmetically:

- **Fallback works as described**: with npm detection empty (simulated by clearing `CLAUDE_CODE_VERSION` and driving the fallback branch directly), `claude --version` output (`2.1.220 (Claude Code)` on this Homebrew-only machine) is correctly parsed by `grep -oE '^[0-9]+\.[0-9]+\.[0-9]+'` into `2.1.220`.
- **Hard-failure behavior verified**: reproduced the "neither method finds a version" branch in isolation — it prints an `ERROR:` to stderr and `exit 1`s, and setting `ALLOW_STALE_USER_AGENT=1` correctly opts back into proceeding with the existing (possibly stale) User-Agent instead of silently continuing. This matches the issue's requirement to fail loudly rather than proceed silently.
- `bash -n scripts/build-macos-app.sh` and `shellcheck scripts/build-macos-app.sh` are both clean.
- `.claude/commands/release.md` changes are coherent: numbered list renumbered correctly, new "verify the patched User-Agent" step added, and the "Important Notes" section updated to mention the fallback + hard-failure behavior.
- CI: `build-macos` and `build-linux` both pass; `mergeStateStatus` is `CLEAN`.

No issues found. Good, targeted fix with real verification rather than just a description of intended behavior.

PR #48, opened 2026-07-31
macOS: dispatch --version at top level so it never launches the GUI

### LINKED ISSUE #46: macOS: `--version` without `--headless` launches the GUI instead of printing
## Context

During the v1.18.0 release verification, running the macOS binary as `ClaudeMonitor --version` did not print the version — it launched a full second menubar instance alongside the installed app (two icons in the menu bar until the stray process was killed). On macOS the informational flags are only handled behind `--headless` (`ClaudeMonitor --headless --version` works); `selftest` is dispatched at top level in `main.swift`, but `--version` (and `--once`) are not.

The README/CLAUDE.md flag documentation (`--once`, `--interval <sec>`, `--version`) doesn't mention that these require `--headless` on macOS, so the natural invocation misfires in the worst way for a menubar app: a duplicate GUI instance.

- [ ] `ClaudeMonitor --version` prints the version and exits on macOS, no GUI
- [ ] `ClaudeMonitor --once` (and other headless-loop flags) either work top-level or fail fast with a clear "use --headless" message — anything but launching the GUI
- [ ] Docs updated to match the final behavior

---

### Root Cause

`menubar-app/ClaudeMonitor/Sources/main.swift`'s `ClaudeMonitorEntry.main()` dispatches on `CommandLine.arguments` in this order today:

```swift
if CommandLine.arguments.dropFirst().first == "accounts" {
    AccountSyncCLI.main(...)
} else if CommandLine.arguments.dropFirst().first == "codex" {
    CodexCLI.main(...)
} else if CommandLine.arguments.dropFirst().first == "selftest" {
    SelfTest.main(...)
} else if CommandLine.arguments.contains("--headless") {
    HeadlessRunner.main()
} else {
    ClaudeMonitorApp.main()   // <- --version / --once / --interval fall through to here
}
```

`--version`, `--once`, and `--interval <sec>` are only recognized **inside** `HeadlessRunner.main()` (see `menubar-app/ClaudeMonitor/Sources/HeadlessRunner.swift:16-36`, which does its own `args.contains("--version")` / `args.contains("--once")` / `args.firstIndex(of: "--interval")` parsing). Since that code only runs when `--headless` is also present, any of these three flags used bare on macOS falls through the final `else` and launches the full `ClaudeMonitorApp` (SwiftUI menubar app) — which is how the v1.18.0 release verification ended up with two menubar icons.

This mirrors the pattern already used for `selftest`/`accounts`/`codex`, which **are** dispatched at top level before the `--headless`/GUI branch — `--version` (and friends) simply never got the same top-level treatment when `HeadlessRunner` grew them.

Linux's entry point (`HeadlessMain.swift`) doesn't have this problem: there is no GUI branch to fall into, so `--version`/`--once` always reach `HeadlessRunner.main()` regardless of `--headless`.

### Implementation Guidance

Add a top-level `--version` case to `ClaudeMonitorEntry.main()` in `main.swift`, ordered *before* the `--headless`/GUI `else`-chain (consistent with how `selftest` is already special-cased), so it never depends on `--headless` being present:

```swift
} else if CommandLine.arguments.contains("--version") {
    print("claude-monitor \(AppVersion.current)")
    exit(0)
} else if CommandLine.arguments.contains("--headless") {
    HeadlessRunner.main()
} else if CommandLine.arguments.contains("--once") || CommandLine.arguments.contains("--interval") {
    FileHandle.standardError.write(Data("--once/--interval require --headless on macOS, e.g. `ClaudeMonitor --headless --once`\n".utf8))
    exit(2)
} else {
    ClaudeMonitorApp.main()
}
```

This satisfies both suggested acceptance criteria in one change:
- `--version` works identically whether or not `--headless` is present, and never launches the GUI.
- `--once`/`--interval` fail fast with a clear stderr message and non-zero exit (matching the existing `--interval` validation-failure convention at `HeadlessRunner.swift:31-34`, which also exits `2`) instead of silently spawning a duplicate GUI instance.

An alternative considered: silently *imply* `--headless` when `--once`/`--interval` are passed bare, so they "just work" top-level. Rejected in favor of the fail-fast message — implying a mode change from a flag combination is more surprising than an explicit error, and the fail-fast path is a one-line diff versus threading `once`/`pollInterval` state into the GUI dispatch path.

`--help`/`-h` have the same latent gap (also only handled inside `HeadlessRunner`) but are not in this issue's acceptance criteria; worth a one-line follow-up but out of scope here to keep the diff focused.

### Affected Files

- `menubar-app/ClaudeMonitor/Sources/main.swift` - add top-level `--version` handling (print + exit) and a fail-fast `--once`/`--interval` guard, both ordered before the `--headless`/GUI dispatch, in `ClaudeMonitorEntry.main()`
- `README.md` - the "Run" section (around line 383-402) lists `--once`/`--interval`/`--version` under the Linux daemon usage and only notes "`ClaudeMonitor --headless`" separately; clarify that on macOS, `--version` now works bare while `--once`/`--interval` still require `--headless` (and will error clearly if not)
- `menubar-app/ClaudeMonitor/Sources/HeadlessRunner.swift` - reference only, no changes expected; confirms the existing `--version`/`--once`/`--interval` parsing this issue must not duplicate or diverge from

### Test Plan

- [ ] Manual verification (macOS, built app or bare binary): `ClaudeMonitor --version` prints `claude-monitor <version>` and exits immediately — confirm via `ps`/Activity Monitor that no new menubar process / status item appears
- [ ] Manual verification: `ClaudeMonitor --once` (bare, no `--headless`) prints a clear stderr error and exits non-zero, with no GUI launch
- [ ] Manual verification: `ClaudeMonitor --interval 300` (bare) behaves the same as `--once` above
- [ ] Manual verification: `ClaudeMonitor --headless --version`, `ClaudeMonitor --headless --once`, and `ClaudeMonitor --headless --interval 300` continue to work exactly as before (no regression to the existing `HeadlessRunner` behavior)
- [ ] Regression: `ClaudeMonitor` (no args) still launches the GUI as normal; `ClaudeMonitor selftest`, `ClaudeMonitor accounts ...`, `ClaudeMonitor codex ...` still dispatch as before
- [ ] Automated tests: N/A — `ClaudeMonitorEntry.main()` is the `#if os(macOS)`-fenced `@main` entry point and isn't unit-testable via `SelfTest`; verification is manual per above. (If a future refactor extracts the dispatch logic into a portable, testable function, that would be a good follow-up but is out of scope here.)
- [ ] Docs: `README.md` "Run" section reads correctly after the change and doesn't contradict the new macOS behavior

### Note on process history

This issue reached Curator carrying only `loom:building` (applied by loom-daemon) with no `loom:curated`/`loom:issue` in its label history — it appears to have been claimed for building without going through the normal Curator → Approval pipeline. This curation pass is being run out-of-band to backfill that gap before the sweep's Approval gate promotes it. Per curator.md, `loom:building` is left untouched here; promotion/label reconciliation is the sweep orchestrator's concern, not Curator's.

rjwalters:
## Summary

`ClaudeMonitor --version` on macOS fell through to `ClaudeMonitorApp.main()` (the SwiftUI menubar app) instead of printing the version, because `--version`/`--once`/`--interval` were only parsed inside `HeadlessRunner.main()`, which is only reached when `--headless` is also passed. `selftest`/`accounts`/`codex` were already dispatched at top level in `main.swift`; `--version` never got the same treatment when `HeadlessRunner` grew it.

## Fix

In `menubar-app/ClaudeMonitor/Sources/main.swift`'s `ClaudeMonitorEntry.main()`, ordered before the `--headless`/GUI dispatch:

- `--version` now prints `claude-monitor <version>` and exits `0` unconditionally (whether or not `--headless` is present) — it never reaches the GUI branch.
- Bare `--once`/`--interval` (without `--headless`) now fail fast with a clear stderr message and exit `2`, instead of silently launching a duplicate GUI instance.

Docs updated to match (`README.md` "Run" section, `CLAUDE.md` entry-points note).

## Test plan

- [x] `swift build` succeeds (no new warnings/errors introduced)
- [x] `swift build && .build/debug/ClaudeMonitor selftest` — 126 checks passed
- [x] `ClaudeMonitor --version` (bare) prints `claude-monitor 1.18.1` and exits 0, no GUI process spawned
- [x] `ClaudeMonitor --headless --version` unchanged — prints and exits 0
- [x] `ClaudeMonitor --once` (bare) prints `--once/--interval require --headless on macOS, e.g. \`ClaudeMonitor --headless --once\`` to stderr and exits 2
- [x] `ClaudeMonitor --interval 300` (bare) — same fail-fast behavior, exit 2
- [x] `ClaudeMonitor --headless --once` — full poll cycle runs and exits 0 (regression check, unchanged)
- [x] `ClaudeMonitor selftest --help`, `ClaudeMonitor accounts --help`, `ClaudeMonitor codex --help` still dispatch correctly (regression check, unchanged)

Closes #46

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,7 +17,7 @@
 
 - The same package builds on Linux (`swift build` with `libsqlite3-dev` installed) as a headless daemon for Loom hosts — no UI, same poll loop, same `usage.db`/`ranking.json` outputs. See README "Headless Mode / Linux".
 - UI sources are fenced with `#if os(macOS)`; portable core must stay free of AppKit/SwiftUI/Combine/os.Logger. On Linux, `LinuxCompat.swift` shims `ObservableObject`/`@Published`, and `CSQLite/` maps the system libsqlite3 (macOS uses the SDK's `SQLite3` module).
-- Entry points: `main.swift` (macOS, dispatches to `HeadlessRunner` on `--headless`) and `HeadlessMain.swift` (Linux, always headless). Flags: `--once`, `--interval <sec>`, `--version`. Subcommands: `accounts`, `codex`, `selftest`.
+- Entry points: `main.swift` (macOS, dispatches to `HeadlessRunner` on `--headless`) and `HeadlessMain.swift` (Linux, always headless). Flags: `--once`, `--interval <sec>`, `--version`. Subcommands: `accounts`, `codex`, `selftest`. On macOS, `--version` is dispatched top-level in `main.swift` (works with or without `--headless`, never launches the GUI); bare `--once`/`--interval` (without `--headless`) fail fast with a stderr message instead of falling through to the GUI branch.
 - Run the test suite with `swift build && .build/debug/ClaudeMonitor selftest` (exits non-zero on failure). CI runs it on macOS and Linux.
 - To verify a schema migration against real data, copy the live DB first — `cp ~/.claude-monitor/usage.db /tmp/ && .build/debug/ClaudeMonitor selftest --db /tmp/usage.db`. `--db` **writes** to the path it is given; never point it at `~/.claude-monitor/usage.db`.
 - Parse response headers via `extractAnthropicHeaders` (lowercased map), never `allHeaderFields` subscripts — those are case-sensitive on Linux.
--- a/README.md
+++ b/README.md
@@ -400,6 +400,11 @@ while the daemon runs. A sample systemd user unit is provided at
 
 On macOS the same headless loop is available as `ClaudeMonitor --headless`
 (the bare binary or the app bundle's `Contents/MacOS/ClaudeMonitor`).
+`ClaudeMonitor --version` prints the version and exits on macOS with or
+without `--headless` — it never launches the GUI. `--once` and `--interval`
+are headless-loop flags: bare (without `--headless`) they print an error to
+stderr and exit non-zero rather than launching a duplicate GUI instance, e.g.
+`ClaudeMonitor --headless --once`.
 
 ## Multi-Host Sync
 
--- a/menubar-app/ClaudeMonitor/Sources/main.swift
+++ b/menubar-app/ClaudeMonitor/Sources/main.swift
@@ -60,8 +60,20 @@ enum ClaudeMonitorEntry {
             CodexCLI.main(Array(CommandLine.arguments.dropFirst(2)))
         } else if CommandLine.arguments.dropFirst().first == "selftest" {
             SelfTest.main(Array(CommandLine.arguments.dropFirst(2)))
+        } else if CommandLine.arguments.contains("--version") {
+            // Dispatched at top level (not just inside HeadlessRunner) so
+            // `ClaudeMonitor --version` never falls through to the GUI branch
+            // below — see #46.
+            print("claude-monitor \(AppVersion.current)")
+            exit(0)
         } else if CommandLine.arguments.contains("--headless") {
             HeadlessRunner.main()
+        } else if CommandLine.arguments.contains("--once") || CommandLine.arguments.contains("--interval") {
+            // These are headless-loop flags handled inside HeadlessRunner; bare
+            // (without --headless) they must fail fast rather than silently
+            // launching a duplicate GUI instance — see #46.
+            FileHandle.standardError.write(Data("--once/--interval require --headless on macOS, e.g. `ClaudeMonitor --headless --once`\n".utf8))
+            exit(2)
         } else {
             ClaudeMonitorApp.main()
         }


rjwalters:
Approved. Verified locally in the builder worktree (`.loom/worktrees/issue-46`):

- `swift build` succeeds.
- `.build/debug/ClaudeMonitor selftest` → `126 check(s) passed`.
- `ClaudeMonitor --version` (bare) → prints `claude-monitor 1.18.1`, exit 0, no GUI.
- `ClaudeMonitor --headless --version` → same output, exit 0 (unchanged path via HeadlessRunner's own `--version` handling, now shadowed by the top-level check but producing identical output).
- `ClaudeMonitor --once` (bare) and `ClaudeMonitor --interval 300` (bare) → stderr message `--once/--interval require --headless on macOS, e.g. \`ClaudeMonitor --headless --once\``, exit 2.
- `ClaudeMonitor selftest --help` → still dispatches correctly, unaffected.

Code review:
- Dispatch order in `ClaudeMonitorEntry.main()` is correct: subcommands (`accounts`/`codex`/`selftest`) → `--version` → `--headless` → bare `--once`/`--interval` guard → GUI fallback. `--version` is checked before `--headless`, so it short-circuits regardless of whether `--headless` is also passed, and it can never fall through to `ClaudeMonitorApp.main()`.
- `exit(2)` for the new fail-fast path matches the project's existing convention for CLI usage errors (`AccountSyncCLI.swift`, `CodexCLI.swift`, `HeadlessRunner.swift`'s own `--interval` validation all use `exit(2)`).
- Change is confined to `main.swift`'s macOS-only entry-point dispatch — no portable-core files touched, so the `#if os(macOS)` / AppKit-SwiftUI boundary isn't implicated here.
- README.md and CLAUDE.md updates accurately describe the new behavior.
- Small, surgical diff (12 lines in main.swift) with no regressions to existing subcommand routing.

CI: both `build-macos` and `build-linux` pass. `mergeStateStatus` is `CLEAN`. PR body includes `Closes #46`.

LGTM.

PR #49, opened 2026-07-31
fix: provider-aware column titles for premium/extra summary columns

### LINKED ISSUE #47: Provider-aware column titles in the summary table ("Fable Left" is meaningless for OpenAI rows)
## Context

The summary-table popover has one shared header row (`UsagePopoverView.swift:610-637`) over mixed-provider rows. Two columns are Anthropic-premium concepts — **Fable Left** and **Extra** — and OpenAI rows show "—" under headings that mean nothing for them. With more providers coming (e.g. Gemini someday), per-provider column naming should be a first-class concept rather than more special cases.

Design constraint: the single mixed table is load-bearing. Cross-provider headroom ranking ("which account should I use right now") depends on all rows sorting together under one column set — so per-provider *sections* with their own headers would break the app's core answer. Retitling based on a selected row is also misleading while the column still renders every provider's data.

## Suggested direction

- **Per-provider column vocabulary**: each provider declares its title for each column slot and which slots apply. Anthropic: premium slot = "Fable Left". OpenAI: slot not applicable — or repurposed to summarize its per-model `named_limits` data, which is already persisted. A future Gemini provider just adds a vocabulary entry.
- **Specific when possible, neutral when mixed**: if every visible row is one provider, show that provider's titles; a mixed table shows a neutral title (e.g. "Premium Left") with the per-provider meaning in the column tooltip (one already exists at `UsagePopoverView.swift:994`).
- **Optional provider filter** in the footer — this is what makes "titles based on what's selected" coherent: filter to one provider and its titles (and only its applicable columns) can take over.

- [ ] No column heading presents an Anthropic-only concept over an OpenAI row without a neutral/mixed-mode treatment
- [ ] Column titles/applicability come from a per-provider vocabulary, not inline conditionals, so adding a provider touches one mapping
- [ ] Cross-provider headroom sorting in the mixed table is unchanged
- [ ] Tooltips explain the per-provider meaning wherever a neutral title is shown

rjwalters:
## Summary
The summary-table popover shared one header row for all providers, so "Fable Left" and "Extra" (both Anthropic-premium concepts) were shown over OpenAI rows that render "—" under them. This adds a per-provider column vocabulary so each provider declares its title (and applicability) for these slots, falling back to a neutral title with an explanatory tooltip whenever the table isn't a single applicable provider.

## Changes
- Added `ProviderColumnSlot` (`.premium`, `.extra`) and `ProviderColumnEntry` (title / applicability / tooltip meaning) in `UsagePopoverView.swift`.
- Added `AccountProvider.columnEntry(for:)` — the single per-provider mapping a future provider (e.g. Gemini) needs to touch to declare its own title for these slots. Anthropic keeps its existing "Fable Left" / "Extra" titles; OpenAI declares both slots not applicable.
- Added `columnHeading(for:neutralTitle:visibleProviders:)`: if exactly one provider is visible in the table and it declares the slot applicable, use its specific title/tooltip (Anthropic-only or OpenAI-only tables are unchanged); otherwise fall back to a neutral title ("Premium Left" / "Extra") whose tooltip explains every provider's meaning for the slot.
- `SummaryHeaderRow` now takes the set of providers currently visible (`UsagePopoverView.visibleProviders`, derived from `store.accounts`) and threads the computed title/tooltip into `SortableHeader`, which gained an optional `tooltip` override (defaults to the existing "Sort by <title>" behavior for every other column).
- No changes to row rendering, sort keys/values, or cross-provider headroom scoring — only header titles/tooltips are affected.

|-----------|--------|--------------|
| No column heading presents an Anthropic-only concept over an OpenAI row without a neutral/mixed-mode treatment | ✅ | A mixed (or OpenAI-only) table now shows "Premium Left" instead of "Fable Left"; an Anthropic-only table is unchanged |
| Column titles/applicability come from a per-provider vocabulary, not inline conditionals | ✅ | `AccountProvider.columnEntry(for:)` is the single mapping; header/row views call `columnHeading(...)`, no per-provider `if`/`switch` in the view body |
| Cross-provider headroom sorting in the mixed table is unchanged | ✅ | `sortValues`/`compareRows`/`headroomScore` untouched — diff is confined to header title/tooltip plumbing |
| Tooltips explain the per-provider meaning wherever a neutral title is shown | ✅ | `columnHeading` builds the neutral-mode tooltip by joining every provider's `meaning` string when it falls back |

## Test Plan
- `swift build` — clean build, no new warnings/errors introduced (pre-existing warnings in `AnthropicAPI.swift` unrelated to this change).
- `.build/debug/ClaudeMonitor selftest` — 126/126 checks passed.
- Manually traced: Anthropic-only accounts → header unchanged ("Fable Left" / "Extra"); mixed Anthropic+OpenAI accounts → header falls back to "Premium Left" with a tooltip describing both providers' meaning; OpenAI-only accounts → same neutral fallback (OpenAI declares the slot not applicable).

## Pre-existing Issues (Not Addressed)
`swift build -Xswiftc -swift-version -Xswiftc 6` (the stronger Swift-6-mode proxy CLAUDE.md recommends) fails with ~65 pre-existing strict-concurrency errors across the codebase (`FileLogger`, `UpdateChecker`, `SelfTest`, chart/roll-token window dictionaries, etc.), confirmed present on `main` before this change and unrelated to `UsagePopoverView.swift`. Not addressed here — out of scope for this issue.

Closes #47

### DIFF
--- a/menubar-app/ClaudeMonitor/Sources/UsagePopoverView.swift
+++ b/menubar-app/ClaudeMonitor/Sources/UsagePopoverView.swift
@@ -52,6 +52,90 @@ enum SummaryColumns {
     static let horizontalPadding: CGFloat = 12
 }
 
+// MARK: - Provider column vocabulary
+
+/// A column whose *meaning* is provider-specific — today "Fable Left" and
+/// "Extra" are both Anthropic-premium concepts that mean nothing for an
+/// OpenAI row. Adding a provider that has something to say for a slot means
+/// adding one case to `AccountProvider.columnEntry(for:)`; no conditionals
+/// scattered through the header/row views.
+enum ProviderColumnSlot {
+    /// Anthropic: "Fable Left" — remaining premium-model weekly allowance.
+    case premium
+    /// Anthropic: "Extra" — overage/extra-usage balance beyond that allowance.
+    case extra
+}
+
+/// One provider's declared title (and whether the slot applies to it at all)
+/// for a `ProviderColumnSlot`.
+struct ProviderColumnEntry {
+    let title: String
+    let isApplicable: Bool
+    /// This provider's meaning for the slot, folded into the neutral-title
+    /// tooltip whenever a mixed (or not-applicable) table falls back to it.
+    let meaning: String
+}
+
+extension AccountProvider {
+    /// This provider's vocabulary entry for a given column slot. The single
+    /// mapping every provider (including a future Gemini client) needs to
+    /// touch to make the summary table's headers make sense for its rows.
+    func columnEntry(for slot: ProviderColumnSlot) -> ProviderColumnEntry {
+        switch (self, slot) {
+        case (.anthropic, .premium):
+            return ProviderColumnEntry(
+                title: "Fable Left",
+                isApplicable: true,
+                meaning: "Anthropic: Fable/premium weekly allowance remaining. At 0% the account switches to extra usage."
+            )
+        case (.anthropic, .extra):
+            return ProviderColumnEntry(
+                title: "Extra",
+                isApplicable: true,
+                meaning: "Anthropic: extra usage (overage) balance beyond the premium allowance."
+            )
+        case (.openai, .premium):
+            return ProviderColumnEntry(
+                title: "Premium",
+                isApplicable: false,
+                meaning: "OpenAI/Codex: no premium-allowance concept — not applicable."
+            )
+        case (.openai, .extra):
+            return ProviderColumnEntry(
+                title: "Extra",
+                isApplicable: false,
+                meaning: "OpenAI/Codex: no extra-usage concept — not applicable."
+            )
+        }
+    }
+}
+
+/// Header title + tooltip for a provider-specific slot, given the providers
+/// actually visible in the table right now. When exactly one provider is
+/// visible *and* it declares the slot applicable, its own vocabulary wins
+/// (unchanged behavior for an Anthropic-only or OpenAI-only table); anything
+/// else — a mixed table, or a lone provider with nothing to say — falls back
+/// to a neutral title whose tooltip spells out every provider's meaning, so
+/// the mixed-table case never presents one provider's concept as if it
+/// applied to all rows.
+func columnHeading(
+    for slot: ProviderColumnSlot,
+    neutralTitle: String,
+    visibleProviders: Set<AccountProvider>
+) -> (title: String, tooltip: String) {
+    if visibleProviders.count == 1,
+       let only = visibleProviders.first {
+        let entry = only.columnEntry(for: slot)
+        if entry.isApplicable {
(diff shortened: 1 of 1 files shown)


rjwalters:
Approved. Verified against issue #47's acceptance criteria:

- **No Anthropic-only concept over OpenAI rows without neutral treatment**: `columnHeading(for:neutralTitle:visibleProviders:)` only uses a provider's own title when exactly one provider is visible *and* that provider declares the slot applicable. A mixed table, or an OpenAI-only table (OpenAI declares both `.premium`/`.extra` as `isApplicable: false`), falls back to the neutral "Premium Left"/"Extra" title. Confirmed by reading the logic and diff — no path renders "Fable Left" over a non-Anthropic-only table.
- **Per-provider vocabulary, not inline conditionals**: `AccountProvider.columnEntry(for:)` is the single switch statement mapping (provider, slot) → title/applicability/meaning. The view body (`SummaryHeaderRow`) has zero per-provider conditionals — it just calls `columnHeading` and threads the result into `SortableHeader`. Adding Gemini would mean adding two cases to this one switch.
- **Cross-provider headroom sorting unchanged**: diff is confined to `UsagePopoverView.swift`'s header/tooltip plumbing (new `ProviderColumnSlot`/`ProviderColumnEntry`/`columnHeading`, `visibleProviders`, and the two `SortableHeader` call sites for fable/extra). No changes to `sortValues`, `compareRows`, `headroomScore`, or row rendering — confirmed via `git diff`.
- **Tooltips explain per-provider meaning on neutral titles**: the fallback path in `columnHeading` joins `AccountProvider.allCases.map { $0.columnEntry(for: slot).meaning }`, so the neutral-title tooltip always spells out what every provider means by the slot.

**Build/test verification** (in the existing `.loom/worktrees/issue-47` builder worktree):
- `swift build` — clean, `.build/debug/ClaudeMonitor selftest` → 126/126 checks passed.
- `swift build -Xswiftc -swift-version -Xswiftc 6` (the strict-concurrency proxy from CLAUDE.md) does surface 4 `oauthPoller` sending-risk errors in `UsagePopoverView.swift`, but `git blame` confirms all 4 sites predate this PR (commits from 2026-02-09 through 2026-07-30) and are untouched by this diff — consistent with the PR description's "Pre-existing Issues" note. No new strict-concurrency errors introduced by the added code.
- CI: `build-linux` initially failed on a transient Docker Hub registry timeout pulling `swift:6.1` (`context deadline exceeded` / `net/http: request canceled`), unrelated to the code — reran the job and it passed cleanly. Both `build-macos` and `build-linux` now pass.
- `mergeStateStatus`: `CLEAN`.

Nice, tight diff — the `ProviderColumnEntry`/`columnEntry(for:)` design is exactly the "touch one mapping" shape the issue asked for.

PR #50, opened 2026-07-31
fix: heal duplicate provider accounts sharing an email at launch

### LINKED ISSUE #45: Healing migration: merge duplicate provider accounts sharing an email
## Context

10660f3 (v1.18.0) stops `codex import` from creating a duplicate account row when an OpenAI account predating the native-id era is keyed by a locally generated UUID: the import now resolves the target row by email within the provider before falling back to OpenAI's native `user-…` id.

That fix is prevention only. Any database where the duplicate already exists keeps two active rows for the same account, each with its own credential, both polled independently. One such DB was repaired by hand this session:

- move `usage_history`, `probe_snapshots`, and `named_limits` rows onto the surviving id (the native `user-…` id, since future imports target it)
- keep exactly one credential (the most recently renewed), reassigned to the surviving row
- delete the sibling account row

Other installs (e.g. headless Loom hosts) may carry the same duplication with nobody watching the menubar to notice.

- [ ] A one-time healing migration at launch (same pattern as the #23 `email=NULL` healing) detects account pairs with the same email and same provider
- [ ] History tables (`usage_history`, `probe_snapshots`, `named_limits`) are merged onto one surviving row; prefer the provider-native id when one of the pair has it
- [ ] Exactly one credential survives (most recently renewed); the sibling account row and its credential are removed
- [ ] Idempotent: re-running on a healed DB is a no-op
- [ ] Covered by a selftest over a scratch DB

rjwalters:
## Summary

- `10660f3` (v1.18.0) stopped a **fresh** `codex import` from forking a duplicate account row for an OpenAI account whose original row predated the native-id era (keyed by a locally generated UUID). That fix is prevention only — a database where the duplicate already exists keeps both rows polling the same account independently, each with its own credential.
- Adds a one-time-per-launch healing migration, following the existing `#23` `email=NULL` healing pattern in `UsageStore.applySchema`: account pairs sharing `(email, provider)` are merged into one surviving row.

## Changes

- `menubar-app/ClaudeMonitor/Sources/UsageStore.swift`: `mergeDuplicateAccountsSharingEmail(_:)`, called from `applySchema` after the email backfill. For each `(email, provider)` group with more than one row:
  - `pickMergeSurvivor` prefers a provider-native id (one that doesn't parse as a canonical UUID — the shape Swift's `UUID()` produces, and the shape the pre-native-id-era duplicate is keyed by) over a locally generated one; ties break on the most-recently-updated row.
  - `mergeAccountRow` moves `usage_history`, `probe_snapshots`, and `named_limits` rows onto the survivor, keeps exactly one `oauth_credentials` row (the more recently renewed of the pair, by `COALESCE(token_rolled_at, updated_at)`), repoints the `primary_account_id` settings pin if it followed the row being removed, and deletes the losing account row — all inside a transaction so a mid-merge failure can't split history across two rows.
- `menubar-app/ClaudeMonitor/Sources/SelfTest.swift`: `testMergeDuplicateAccountsSharingEmail()` over a scratch DB — covers the merge itself, provider-scoping (an Anthropic row sharing the email is untouched), the credential-recency tie-break, the settings pin repoint, and idempotency (re-running the healed DB is a no-op).
- `CHANGELOG.md`: entry under the current `[1.18.1]` section (matching how `10660f3` recorded its fix), following this repo's existing practice of appending to the latest section between releases.

- [x] One-time healing migration at launch, same pattern as the `#23` `email=NULL` healing — `mergeDuplicateAccountsSharingEmail` runs unconditionally from `UsageStore.applySchema`, called by `ensureDatabase()` on every launch.
- [x] `usage_history`, `probe_snapshots`, `named_limits` merged onto one surviving row; provider-native id preferred when one of the pair has it — verified by `testMergeDuplicateAccountsSharingEmail`.
- [x] Exactly one credential survives (most recently renewed); sibling row and its credential removed — verified (`legacy-fresher-token` wins over a newer-id row with a staler `token_rolled_at`).
- [x] Idempotent: re-running on a healed DB is a no-op — verified by a second `ensureDatabase()` call in the test asserting unchanged row counts.
- [x] Covered by a selftest over a scratch DB — `testMergeDuplicateAccountsSharingEmail`, registered in `SelfTest.main`.

## Test Plan

- `swift build && .build/debug/ClaudeMonitor selftest` — `selftest: 139 check(s) passed` (was 138 before this change; +1 new test).
- Linux build verified via the `swift:6.1` Docker image (`apt-get install libsqlite3-dev`, `swift build`, `selftest`) — build succeeds, all 139 checks pass.
- `swift build -Xswiftc -swift-version -Xswiftc 6` (the stronger local proxy for CI's strict-concurrency checks) was run — it reports the same pre-existing Swift-6-language-mode errors already present on `main` (unrelated `FileLogger`/`UpdateChecker`/`SelfTest` static-state warnings across the codebase, confirmed identical error count on `main` before this change), so this is not a regression introduced here.

Closes #45

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,15 @@ credential when an OpenAI account shares its email address.
   recover the history by reassigning non-wham-shaped `usage_history` rows (and
   `fable`/`haiku` `probe_snapshots`) back to the Anthropic org id and
   re-importing a valid token
+- **Healing migration merges duplicate provider accounts sharing an email.**
+  `10660f3` only stopped a *fresh* `codex import` from forking a second row for
+  an OpenAI account whose original row predated the native-id era (keyed by a
+  locally generated UUID); a database where the duplicate already existed kept
+  both rows polling the same account independently. At launch, account pairs
+  sharing `(email, provider)` are now merged automatically: history moves onto
+  the surviving row (preferring the provider-native id), exactly one
+  credential survives (the more recently renewed of the two), and the sibling
+  row is removed. Idempotent, and covered by a selftest over a scratch DB (#45)
 
 ## [1.18.0] - 2026-07-30
 
--- a/menubar-app/ClaudeMonitor/Sources/SelfTest.swift
+++ b/menubar-app/ClaudeMonitor/Sources/SelfTest.swift
@@ -54,6 +54,7 @@ enum SelfTest {
         testNamedLimitsRoundTrip()
         testOpenAIImportResolvesExistingAccountByEmail()
         testAccountSyncImportIsProviderScoped()
+        testMergeDuplicateAccountsSharingEmail()
 
         if let idx = arguments.firstIndex(of: "--db"), idx + 1 < arguments.count {
             testMigrationOfExistingDatabase(at: arguments[idx + 1])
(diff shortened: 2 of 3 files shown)


rjwalters:
Judge pass: PR still carries a fresh `loom:reviewing` claim (claimed 2026-07-31T08:27:44Z) — standing down without reclaiming. Not stomping.

rjwalters:
Approved.

Reviewed against issue #45's acceptance criteria:

- **One-time healing migration at launch** — `mergeDuplicateAccountsSharingEmail` runs unconditionally from `UsageStore.applySchema` (called by `ensureDatabase()` every launch), after the `#23` email backfill so a row healed there is eligible too. Same pattern as the existing `email=NULL` healing.
- **History tables merged onto the survivor** — `usage_history`, `probe_snapshots`, and `named_limits` are repointed via `UPDATE ... SET account_id = ?`. `pickMergeSurvivor` correctly prefers a provider-native id (non-UUID-shaped) over a locally generated one, tying on most-recently-updated. Verified schema column names (`last_updated`, table names) match the actual `CREATE TABLE` statements.
- **Exactly one credential survives** — the more recently renewed (`COALESCE(token_rolled_at, updated_at)`) is reassigned to the survivor; any other credential row for either id is deleted. Test confirms the *legacy* row's credential wins even though the *native* row's id survives — good, non-obvious case actually exercised.
- **Idempotent** — grouping by `(email, provider)` naturally finds nothing once only one row remains per pair; the test calls `ensureDatabase()` twice and asserts unchanged row counts.
- **Selftest over a scratch DB** — `testMergeDuplicateAccountsSharingEmail` covers the merge, provider-scoping (an Anthropic row sharing the email is untouched), the credential tie-break, the `primary_account_id` settings-pin repoint, and idempotency. Registered in `SelfTest.main`.

Other things I checked:
- Merge runs inside a `BEGIN`/`COMMIT` transaction with `ROLLBACK` on failure (via `try?` so a rollback failure doesn't mask the original error) — a mid-merge crash can't split history across two rows.
- Provider scoping uses `COALESCE(provider, 'anthropic')`, consistent with the existing guards in `resolveOpenAIAccountId` / `AccountSync.importAccount`.
- One-failure-doesn't-block-others: if a single pair's merge fails, it's logged and the outer loop continues to the next group rather than aborting the whole migration.

Verification performed:
- `swift build && .build/debug/ClaudeMonitor selftest` on this branch: `139 check(s) passed` (was 138 on main — +1 new test, no regressions).
- Clean-rebuilt `swift build -Xswiftc -swift-version -Xswiftc 6` (the strong local proxy for CI's strict-concurrency errors) on both this branch and a fresh `main` checkout and diffed the unique `file:line: error:` locations: **16 pre-existing errors on both**, same single error just shifted from `UsageStore.swift:1112` to `:1232` (the new code's 120 inserted lines pushed the line number) — confirms this PR introduces zero new Swift-6-language-mode diagnostics.
- CI: `build-macos` pass, `build-linux` pass (was pending at claim time, now green).
- Merge state: `CLEAN`, `mergeable: MERGEABLE`.

Nice test design overall — the fixture deliberately makes the "wrong" answer look plausible (legacy row has the fresher credential, native row has the fresher `last_updated`) so the test actually exercises the tie-break logic instead of trivially passing either way.

PR #53, opened 2026-08-02
build: adopt Swift 6 language mode package-wide

rjwalters:
## Summary
- `Package.swift` now passes `-swift-version 6` to the `ClaudeMonitor` target (via `swiftSettings`/`unsafeFlags`, keeping `swift-tools-version:5.9` so the manifest stays parseable by any 5.9+ SwiftPM/toolchain) — a plain `swift build` now enforces full Swift 6 strict-concurrency checking on both macOS and Linux, the same checking CI already exercises.
- Fixed every diagnostic that surfaced, favoring the narrowest correct annotation per site rather than a blanket `nonisolated(unsafe)`:
  - `@MainActor`: `UsageStore`, `OAuthPoller` (both `ObservableObject`s only ever driven from the app's single main-thread call path — SwiftUI on macOS, a `Task`-driven loop pumped via `dispatchMain()` in headless/Linux mode), the AppKit window-controller caches (`RollTokenWindowController`, `ChartWindowController`), `AppDelegate`, `UpdateChecker`, and the CLI/headless entry points that drive them (`CodexCLI`, `HeadlessRunner`, `SelfTest`, both platforms' `ClaudeMonitorEntry.main()`).
  - `Sendable`: `FileLogger` (`@unchecked` — internally serialized via its own private `DispatchQueue`; documented), `AnthropicAPIClient` / `OpenAIAPIClient` (genuinely stateless besides an immutable `URLSession`), and the `UsageProviderClient` protocol they conform to.
  - `nonisolated`: pure `db`/argument-only static helpers on `UsageStore`/`OAuthPoller` that got swept into `@MainActor` isolation only because they live inside those now-`@MainActor` types (schema migration, dedup/merge helpers, `resetSeconds`, `resolveOpenAIAccountId`, etc.) — none of them touch `@Published` state.
  - `nonisolated(unsafe)` (documented at each site, not blanket): `SelfTest`'s single-threaded-by-construction test accumulators, and `AnthropicAPIClient`'s log-dedup `Set<String>` (`OAuthPoller` processes credentials sequentially in an `await`-ed loop, never concurrently).
  - `fflush(nil)` in `FileLogger` instead of `fflush(stdout)` — Linux's Glibc overlay (unlike Darwin's) doesn't mark the C global `stdout` `Sendable`, so any reference to it is flagged; `fflush(nil)` (flush all open streams) is the portable equivalent for the "make sure this just-printed line reaches journald/docker logs" intent.
[...]

PR #54, opened 2026-08-04
Order capped accounts by time until they are usable again

rjwalters:
## What

When sorting by Headroom, every exhausted account reads 0 — so the whole capped block fell through to the display-name tiebreak, and the account coming back in ten minutes sorted next to one that is out for three days.

Ties on the Headroom column now break on the reset that **actually gates** the account:

- weekly window exhausted (100% used, or the provider says `rejected`) → the **weekly** reset gates it
- otherwise → the **session** reset gates it
- no session window at all (an OpenAI reading may have none) → falls back to weekly
- no reset time known → stays `nil` and ranks last, rather than reading as "available now"

## Where it applies

`RateLimitSnapshot` gains `gatingWindow` / `secondsUntilRecovery` in the portable core, so the headless Linux daemon reasons about recovery the same way the popover does.

`UsageStore.resetSeconds` always preferred the session reset; it is now a thin wrapper over `secondsUntilRecovery`. That fixes the identical flaw in `sortedAccountsForPopover`, and therefore in the auto-picked menubar account — an account with a session reset ten minutes out but no weekly capacity used to outrank one that was genuinely two hours from usable.

The popover tiebreak respects sort direction, so flipping the header inverts the recovery order too instead of pinning one end of it.

## Verification

- `swift build` clean under the Swift 6 language mode adopted in #53
- `selftest` passes 144 checks; `testGatingResetOrdering` covers the three gating cases directly, then orders three capped accounts end-to-end through `sortedAccountsForPopover`
- The end-to-end case gives the weekly-exhausted account the **soonest** session reset on purpose, so a regression to the old behavior sorts it first and fails
- Mutation-checked: deleting the weekly-gating line fails exactly those two assertions and nothing else

## Second commit (unrelated, repo hygiene)

`npm test` claimed there was no Swift test target and exited 0 unconditionally — that predates `selftest`, which CI has run on both platforms since #30, so every orchestration gate shelling out to `npm run test` / `check:all` was passing vacuously. It now builds and runs the suite. Also corrects the README's CI description (two jobs, both running selftest) and adds the missing `CLAUDE.md` to the project tree.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   spent, the session reset otherwise. The same rule replaces the old
   session-reset-first tiebreak used for popover ordering and the auto-selected
   menubar account.
+- **`npm test` runs the self-test.** The script still claimed there was no test
+  target and exited 0 unconditionally, which predates `selftest` — orchestration
+  gates that shell out to it now actually gate.
 
 ## [1.18.3] - 2026-08-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- **Capped accounts sort by when they come back.** Accounts tied on headroom —
+  in practice the block of exhausted accounts all reading 0% — now order by time
+  until the reset that actually gates them: the weekly reset once the week is
+  spent, the session reset otherwise. The same rule replaces the old
+  session-reset-first tiebreak used for popover ordering and the auto-selected
+  menubar account.
+
 ## [1.18.3] - 2026-08-02
 
 ### Summary
--- a/README.md
+++ b/README.md
@@ -342,7 +342,9 @@ swift build
 .build/debug/ClaudeMonitor &
 ```
 
-CI (`.github/workflows/build.yml`) runs the same `swift build` on every push.
+CI (`.github/workflows/build.yml`) runs on pushes to `main` and on pull requests:
+two jobs build the package — macOS and a `swift:6.1` Linux container — and each
+runs `ClaudeMonitor selftest`, with the Linux job also smoke-running `--once`.
 
 ### Build for Distribution
 
@@ -651,13 +653,14 @@ claude-monitor/
 │   ├── build-macos-app.sh          # macOS release build script
 │   └── claude-monitor.service      # Sample systemd user unit for Linux headless mode
 ├── docs/spikes/                 # Investigation write-ups (e.g. the OpenAI usage-endpoint probe)
-├── .github/workflows/build.yml  # CI: swift build on push
+├── .github/workflows/build.yml  # CI: build + selftest on macOS and Linux
 ├── build/                       # Build output (gitignored): ClaudeMonitor.app + .zip
-├── CHANGELOG.md
+├── CHANGELOG.md                 # Release history
+├── CLAUDE.md                    # Development notes (build/install sequence, invariants)
 ├── .env.example                 # Sample accounts.env for bulk import
 ├── window.png, plot_window.png  # README screenshots
 ├── loom.sh, package.json        # Loom orchestration workspace files (not part of the app)
-└── .loom/, .claude/             # Loom + Claude Code tooling installs
+└── .loom/, .claude/, .gitattributes  # Loom + Claude Code tooling installs
 ```
 
 ## Related Projects
--- a/menubar-app/ClaudeMonitor/Sources/RateLimitWindow.swift
+++ b/menubar-app/ClaudeMonitor/Sources/RateLimitWindow.swift
@@ -260,4 +260,22 @@ struct RateLimitSnapshot {
     var nextReset: Date? {
         [session?.resetAt, weekly?.resetAt].compactMap { $0 }.min()
     }
+
+    /// The window whose reset actually gates this account's next usable moment.
+    ///
+    /// An exhausted weekly window is the gate: the session limit rolling over in
+    /// twenty minutes buys nothing while the week is spent. Otherwise the
+    /// session window is what frees up first, and an account with no session
+    /// window (an OpenAI reading may have none) falls back to its weekly one.
+    var gatingWindow: RateLimitWindow? {
+        if let weekly = weekly, weekly.isExhausted { return weekly }
+        return session ?? weekly
+    }
+
+    /// Seconds until this account is expected to have capacity again, measured
+    /// against `gatingWindow`. Nil when no relevant reset time is known — so
+    /// callers rank it as *unknown* rather than *imminent*.
+    var secondsUntilRecovery: TimeInterval? {
+        gatingWindow?.resetAt?.timeIntervalSinceNow
+    }
 }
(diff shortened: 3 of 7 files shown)


rjwalters:
✅ **Looks good — code evaluation feedback** (fallback mode, no `loom:` label on this PR, so I'm not touching labels)

Reviewed the change end-to-end and it holds up:

- **Logic**: `RateLimitSnapshot.gatingWindow` correctly prefers the weekly window only when it's actually exhausted (`status == "rejected" || usedPercent >= 100`), falling back to session, then weekly for providers with no session window. `UsageStore.resetSeconds` and the popover's `recoveryTiebreak` both route through `secondsUntilRecovery`, so the headless daemon and the UI now agree.
- **Sort-direction handling**: traced through `recoveryTiebreak` — under the default `.desc` (best-first) sort, ties break soonest-recovery-first; flipping to `.asc` correctly inverts it. Matches the PR description.
- **Build**: clean under plain `swift build` and under the stronger `-swift-version 6` proxy build (per this repo's CLAUDE.md guidance on strict-concurrency false negatives).
- **Tests**: `selftest` passes all 144 checks, including the new `testGatingResetOrdering`.
- **Mutation check**: deleted the `if let weekly = weekly, weekly.isExhausted { return weekly }` line by hand and reran `selftest` — exactly the two claimed assertions failed (`"an exhausted weekly window gates recovery..."` and the end-to-end `sortedAccountsForPopover` ordering test), nothing else regressed. Confirms the test actually exercises the fix rather than passing vacuously.
- **Second commit (repo hygiene)**: verified `npm run test` now actually builds + runs `selftest` (previously `exit 0` unconditionally) — ran it directly, it built and reported 144/144. README/CHANGELOG updates are accurate.
- **CI**: `build-linux` and `build-macos` both pass; merge state is `CLEAN`.

No blocking issues found. Nice, focused change with a genuinely load-bearing test (the mutation check confirms it isn't just decorative).

rjwalters:
✅ **Approved (fallback-mode evaluation — no `loom:review-requested` label present, so no labels changed)**

Reviewed in an isolated worktree (`.loom/worktrees/pr-54`):

- `mergeStateStatus: CLEAN`, both CI jobs (`build-macos`, `build-linux`) pass.
- `swift build && .build/debug/ClaudeMonitor selftest` → **144/144 checks pass**.
- Mutation-tested the core claim: commenting out the weekly-exhaustion check in `RateLimitSnapshot.gatingWindow` (`RateLimitWindow.swift:271`) fails exactly the two assertions the PR description says it should (`testGatingResetOrdering`'s weekly-gate case and the end-to-end `sortedAccountsForPopover` ordering case) and nothing else — 142/144 still pass. Restored cleanly afterward.
- Traced through `recoveryTiebreak`'s direction handling in `UsagePopoverView.swift` by hand against the default (`sortBy: .headroom, sortDir: .desc`) and flipped case — the "sooner recovery sorts first under best-first, and the order inverts when flipped" behavior is correct.
- The `npm test` fix is a real bug fix, not cosmetic — the old script was a no-op `exit 0` regardless of build/test outcome, meaning every orchestration gate shelling out to `npm test`/`check:all` was vacuously green. Verified the new version actually builds and runs `selftest`, and fails non-zero on the induced-failure case above.
- `resetSeconds`/`sortedAccountsForPopover` correctly reduce to `secondsUntilRecovery` with no behavior drift for the non-capped path.

Nice clean PR — good test discipline (a dedicated mutation-style regression case, not just happy-path coverage), and the "second, unrelated hygiene commit" is called out honestly in the description rather than buried.

## Test Execution

PR description has no explicit "## Test Plan" section, but its "Verification" section doubles as one — all three claims (clean Swift 6 build, 144 selftest checks, mutation-kills-exactly-those-two-assertions) were independently reproduced above.

rjwalters:
✅ **Approved!** (Fallback-mode review — no `loom:` labels on this PR, so no label changes made.)

Reviewed the diff in an isolated worktree against current `main` (9119a6b).

**Correctness**
- `RateLimitSnapshot.gatingWindow`/`secondsUntilRecovery` correctly derive the gating window from `weekly.isExhausted` (`status == "rejected" || usedPercent >= 100`), matching the description exactly.
- `UsageStore.resetSeconds` is now a thin wrapper over `secondsUntilRecovery`, fixing the same session-reset-first bug in `sortedAccountsForPopover`.
- The `recoveryTiebreak` sort-direction handling is correct and actually elegant: it keeps the tied (0-headroom) block internally ordered consistently with whichever direction the outer sort is running, so scanning the list stays monotonic in "time until usable" at the boundary between capped and non-capped accounts. Traced through both `.asc`/`.desc` cases by hand — checks out.
- nil-handling is right throughout: unknown recovery sorts last, not "available now".

**Testing**
- `swift build && .build/debug/ClaudeMonitor selftest` → 144/144 pass.
- Also built under `-Xswiftc -swift-version -Xswiftc 6` (the stronger local proxy for CI's strict-concurrency checks per this repo's CLAUDE.md) — clean, only pre-existing unrelated warnings in `AnthropicAPI.swift`.
- Verified the PR's mutation-testing claim myself: reverting `gatingWindow` to drop the weekly-exhausted check fails **exactly** the two assertions the PR says it does (`testGatingResetOrdering`'s weekly-gated check and the end-to-end `sortedAccountsForPopover` ordering) and nothing else — 142/144 pass, 2 fail, both in the new test. Restored afterward.
- Second (hygiene) commit: `npm test` now genuinely builds + runs selftest instead of vacuously exiting 0 — ran it directly, works as described.
- CI (`build-linux`, `build-macos`) both pass; `mergeStateStatus` is `CLEAN`.

**Design**
- Putting `gatingWindow`/`secondsUntilRecovery` on the portable-core `RateLimitSnapshot` (rather than in `UsageStore`) correctly makes the headless Linux daemon share the same recovery reasoning as the popover, consistent with this repo's portable-core/AppKit separation.

No blocking issues found. Nice, well-tested fix with good self-verification in the PR description.

rjwalters:
✅ **Evaluated (fallback mode — no `loom:` labels present, so labels left untouched)**

Solid, well-scoped fix. Verified independently:

- `swift build` clean, including under the Swift 6 language-mode proxy (`-swift-version 6`) called out in `CLAUDE.md` for strict-concurrency changes — no new warnings.
- `selftest` passes 144/144.
- **Mutation-tested the core claim**: deleted the weekly-gating line in `RateLimitSnapshot.gatingWindow` (`RateLimitWindow.swift`) and reran `selftest` — it fails exactly the two assertions the PR description says it should (`testGatingResetOrdering`'s weekly-gate check and the end-to-end `agent-3/agent-1/agent-2` ordering) and nothing else. Restored and confirmed clean again.
- Traced `secondsUntilRecovery`/`resetSeconds`/`gatingWindow` to confirm there's a single source of truth — `UsageStore.resetSeconds`, the popover's `sortedAccountsForPopover`, and the new `recoveryTiebreak` in `UsagePopoverView` all route through the same `RateLimitSnapshot.secondsUntilRecovery`, so there's no duplicated or diverging gating logic left behind.
- Checked the `recoveryTiebreak` sort-direction math (`UsagePopoverView.swift`) by hand against the existing `sortDir == .asc ? (x < y) : (x > y)` convention used elsewhere in the same comparator — it's consistent (sooner recovery leads under "best first" `.desc`, trails when flipped).
- CHANGELOG's `## [Unreleased]` section is new relative to `main` (which has none since the 1.18.3 release), so no duplicate-heading risk.
- Second, unrelated commit fixing `npm test`'s vacuous no-op is a good catch — worth calling out that this had been making every orchestration gate that shells out to `npm run test`/`check:all` pass vacuously.
- CI green on both `build-linux` and `build-macos`; merge state `clean`.

No blocking issues found. This PR doesn't reference a tracking issue (`Closes #N`) — reads as a directly-authored fix rather than a Loom-issue PR, so that's not a defect here.

Issue #102, opened 2026-08-15
Read Codex usage via `codex app-server` instead of storing OAuth tokens


rjwalters:
## Context

`OpenAIAPI.swift` stores an OpenAI access + refresh token in `oauth_credentials` and refreshes it itself against `auth.openai.com/oauth/token`. OpenAI **rotates the refresh token on every use**, so the monitor's stored copy, any second host's copy, and the Codex CLI's own `$CODEX_HOME/auth.json` continually invalidate one another. Whichever client refreshes first silently kills the others.

Observed on two macOS hosts: continuous `[ERROR] [OpenAI] OpenAI token refresh rejected (401/refresh_token_invalidated)` on every poll for ~9 days, with every `provider: openai` row frozen at a stale `last_updated` while Anthropic rows updated normally.

**The accounts were never unhealthy.** Running `codex` 0.147.0 on one of those hosts refreshed `auth.json` in place immediately and returned live usage — only our stored snapshot was dead. We are reporting our own staleness as an account failure. That reframes this from a credential bug to an architectural one: we should not be holding this credential at all.

This is also unsupported by design. OpenAI's own guidance:

> Use one `auth.json` per runner or per serialized workflow stream. Do not share the same file across concurrent jobs or multiple machines.
> — <https://learn.chatgpt.com/docs/auth/ci-cd-auth>

## The supported surface

`codex app-server` speaks JSON-RPC over stdio and returns usage directly, with Codex owning the credential entirely. Verified end to end against `codex-cli 0.147.0`:

```
initialize → initialized (notification) → account/read → account/rateLimits/read
```

`account/read` returns account type, email and plan. `account/rateLimits/read` returns (values redacted):

```json
{"rateLimits": {
   "limitId": "codex",
   "primary": {"usedPercent": <n>, "windowDurationMins": 10080, "resetsAt": <epoch>},
   "secondary": null,
   "credits": {"hasCredits": false, "unlimited": false, "balance": "<n>"},
   "planType": "<plan>",
   "spendControlReached": false},
 "rateLimitsByLimitId": {
   "codex": { ... },
   "codex_bengalfox": {"limitName": "GPT-5.3-Codex-Spark", "primary": {...}}},
 "rateLimitResetCredits": {"availableCount": 0, "credits": []}}
```

Identity, plan, utilization, window duration, reset time, credits **and** per-model sub-limits — with no token read, stored, or refreshed by us.

## Proposal

[...]

Issue #104, opened 2026-08-15
Retire stored OpenAI OAuth tokens and exclude Codex accounts from accounts export


rjwalters:
## Context

Once Codex usage is read via `app-server` with a per-account `CODEX_HOME`, the monitor no longer needs an OpenAI credential of its own. The stored copy stops being merely redundant and becomes actively harmful: every `oauth_credentials` row holding an OpenAI refresh token is a copy of a rotating secret that will invalidate whichever other client refreshes next.

Two clean-up obligations follow, both currently unmet.

**1. Stop persisting OpenAI tokens.** `oauth_credentials` rows for `provider = 'openai'` should no longer carry `access_token` / `refresh_token`, and the proactive-refresh path in `OAuthPoller` should be removed for that provider. Existing rows should be cleared by migration — they are dead secrets sitting in a database that has no reason to hold them.

**2. Stop exporting them.** `accounts export` emits every credential in plaintext, including OpenAI ones. Because a Codex `auth.json` is explicitly single-machine, an exported OpenAI credential is *guaranteed* to break the source host the moment it is imported and refreshed elsewhere. Multi-host sync should carry Anthropic accounts only; a Codex account is registered per host by pointing at a local home, not by shipping a secret between machines.

This is the failure that motivated the redesign — two hosts each holding a copy of the same rotating token, taking turns invalidating each other, with the export/import path being what put the second copy there.

## Proposal

- Migration: null out `access_token` / `refresh_token` for `provider = 'openai'` rows.
- Remove the OpenAI branch of the proactive token-refresh loop in `OAuthPoller`.
- `AccountSync.exportBundle` skips `provider = 'openai'` accounts; `importBundle` ignores any present in an older bundle rather than failing, so bundles produced before this change stay importable.
- `accounts export` notes in its stderr warning that Codex accounts are host-local and excluded by design.
- README "Multi-Host Sync" states that Codex accounts are registered per host.

Depends on #102 and #103 landing first — this is the removal step, and doing it earlier would leave Codex unmonitored.

## Suggested acceptance criteria

- [ ] No code path writes an OpenAI access or refresh token to the database
- [ ] Migration clears existing OpenAI token columns and is covered by `selftest`
- [ ] `accounts export` contains no `provider: openai` account; a bundle from an older version still imports without error
- [ ] Anthropic export/import behaviour is unchanged, including the existing "skip records older than the local row" rule
- [ ] README and `accounts export --help` state that Codex accounts are host-local
[...]

Issue #103, opened 2026-08-15
Support multiple Codex accounts via per-account CODEX_HOME


rjwalters:
> **Curator status banner (2026-08-15).** The "Depends on #102 landing first"
> line in the original body below is **satisfied** — #102 merged as PR #111,
> on `main` at `a46cf07`. This issue branches off plain `main`; there is no
> stacking. Everything below the `---` is the curated spec and supersedes the
> original where they differ.

## Context

`codex login` writes a single `auth.json` per `CODEX_HOME`, so **each login overwrites the previous account's credential**. There is exactly one Codex identity per home directory, and `Sources/CodexCLI.swift` imports from whichever one happens to be there.

The practical consequence: monitoring more than one Codex account has never actually worked. Across two macOS hosts, one held a single `provider: openai` row and the other held two — and a third known Codex account was absent from both, because logging in as it would have destroyed one of the others. Nothing was misconfigured; there was simply nowhere to put it.

`--auth <path>` on `codex import` looks like it solves this, but it only relocates the *read*. The credential still has to be produced by a `codex login` that clobbers the default home, and the imported copy then rots as soon as the token rotates.

## The mechanism

`CODEX_HOME` fully isolates a Codex identity, and the isolation is legible over `app-server`. `initialize` echoes the resolved home back:

```json
{"id":1,"result":{"codexHome":"<path>","platformOs":"macos", ...}}
```

and an unauthenticated home answers `account/read` cleanly rather than failing:

```json
{"id":2,"result":{"account":null,"requiresOpenaiAuth":true}}
```

That gives a per-account "needs login" health signal for free — distinguishable from "logged in but the request failed", which the current single-token model cannot express.

This is the same approach the prior art takes: CodexBar scopes each account by setting `CODEX_HOME` per fetch and lists additional homes in its own config.

## Proposal

Make the Codex home a per-account property rather than a global:

- Add a nullable `codex_home` column to `accounts` (additive migration; NULL means the default `$CODEX_HOME`/`~/.codex`).
- Register an account by its home directory rather than by importing a token — e.g. `claude-monitor codex add --home <path>`, which runs `account/read` against that home and stores the resulting identity.
- `claude-monitor codex list` to show each registered home and whether it is currently authenticated.
- Poll each Codex account with `CODEX_HOME` set to its own directory.

[...]

PR #107, opened 2026-08-15
docs: supersede the Codex spike — app-server replaces stored OAuth tokens

rjwalters:
## Summary

Docs-only. Brings the Codex design documentation in line with what was actually
verified this week, and marks the superseded recommendation as superseded
rather than deleting it — following the convention `docs/spikes/README.md`
states for exactly this case.

The 2026-07-30 spike recommended storing an OpenAI `refresh_token` +
`token_expires_at` and renewing proactively. It deliberately never ran a
successful refresh exchange, so it could not tell that OpenAI rotates the
refresh token on every use. That design shipped, and it does not hold: every
holder of the credential — this app on either host, plus the Codex CLI's own
`$CODEX_HOME/auth.json` — invalidates the others when it renews.

Observed in production: continuous `401 / refresh_token_invalidated` across two
hosts for ~9 days, with every `provider: openai` row frozen, **while the
accounts themselves were entirely healthy**. Running a current `codex`
refreshed `auth.json` in place and returned live usage immediately. The app was
reporting its own staleness as an account failure.

## What changed

- **`docs/spikes/2026-07-30-codex-usage-probe.md`** — new `Supersession
  (2026-08-15)` section, marked authoritative. Records that `codex app-server`
  exposes `account/read` + `account/rateLimits/read` over stdio JSON-RPC with
  Codex owning the credential; that `CODEX_HOME` scopes it per account (and is
  why multi-account never worked — `codex login` writes one `auth.json` per
  home); and the version constraint (absent in codex 0.46.0, verified on
  0.147.0).
- **`docs/spikes/README.md`** — index outcome and reading guide updated for the
  third layer.
- **`README.md`** — the existing rotation note prescribed re-syncing the
  credential between hosts via `accounts export`/`import`. That loop does not
  converge; amended to say so, and to say plainly not to copy OpenAI
  credentials between machines.

## Notes for review

- The live-verification section is **not** contradicted and is explicitly
  called out as still binding — it is now the fallback wire contract, and
  `OpenAIAPI.swift` / `RateLimitWindow.swift` still cite it correctly. Its two
  substantive findings (derive window kind from duration; a session window may
  legitimately be absent) are reinforced by the `app-server` shape, not
  replaced.
- Rotation itself was already documented in the README as verified 2026-07-31.
  What is new here is the *consequence* in sustained multi-host use, and the
  supersession says so rather than claiming a fresh discovery.
- All sample payloads are redacted (`<n>`, `<epoch>`, `<plan>`); no emails,
  hostnames, account ids, or tokens appear in the diff.

Implementation is tracked separately in #102 (app-server client), #103
(per-account `CODEX_HOME`), #104 (retire stored OpenAI tokens). This PR only
stops the docs from recommending the design those issues remove.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,20 @@ What differs from an Anthropic row once it's added:
 >   `accounts export` → `accounts import` (see
 >   [Multi-host sync](#multi-host-sync)),
 >   or just re-run `codex import` there after a fresh `codex login`.
+>
+> **This re-sync loop does not converge, and is being replaced.** Each holder
+> of the credential — this app on host A, this app on host B, the Codex CLI on
+> each of them — invalidates the others when it renews, so re-syncing only
+> moves which one is about to break next. Observed in practice: continuous
+> `401 / refresh_token_invalidated` across two hosts for ~9 days, while the
+> accounts themselves stayed healthy and a current `codex` read their usage
+> immediately. OpenAI's own guidance is that an `auth.json` is single-machine
+> and must not be shared across hosts, so **do not** copy OpenAI credentials
+> between machines; register the account on each host instead. The fix is to
+> stop holding the credential at all and read usage via `codex app-server`,
+> which owns its own auth — see #102, #103 and #104, and the 2026-08-15
+> supersession in
+> [`docs/spikes/2026-07-30-codex-usage-probe.md`](docs/spikes/2026-07-30-codex-usage-probe.md).
 
 ### Rolling a Token (revoke + re-mint)
 
--- a/docs/spikes/2026-07-30-codex-usage-probe.md
+++ b/docs/spikes/2026-07-30-codex-usage-probe.md
@@ -235,7 +235,13 @@ and `Account` schema now — in particular:
   correct without that check — treat it as a strong hypothesis, not a
   verified contract.
 
-## Live verification (2026-07-30) — AUTHORITATIVE
+## Live verification (2026-07-30) — AUTHORITATIVE for the wire contract
+
+> **Partly superseded.** This section's *wire contract* is still correct and is
+> now the fallback path. Its *design* conclusion — store and refresh a
+> credential — was reversed on 2026-08-15; see
+> [Supersession (2026-08-15)](#supersession-2026-08-15--rotation-confirmed-app-server-is-the-supported-surface)
+> at the end of this document before designing against anything here.
 
 The operator judged a disposable account unnecessary: the probe is a
 **read-only GET** against the same endpoint Codex CLI's own `/usage` command
--- a/docs/spikes/2026-07-30-codex-usage-probe.md
+++ b/docs/spikes/2026-07-30-codex-usage-probe.md
@@ -362,6 +362,97 @@ What this establishes, and what it doesn't:
   actually in use. Re-confirmed on 2026-07-30: still `null` on this Pro
   account, with a weekly `primary_window` at 14%.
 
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!**

Documentation-only change (3 files, +122/-9) that brings the Codex design docs in line with what was empirically verified this week. No executing code touched.

## What I verified

- **Diff scope** — `README.md` + the two spike documents only. Confirmed against `origin/main...HEAD`; no source, build, or config files in the diff.
- **The load-bearing claim holds.** `docs/spikes/README.md` asserts that `OpenAIAPI.swift` and `RateLimitWindow.swift` cite the live-verification section. That claim was inherited and restated by this PR, so I checked it rather than trusting it: `OpenAIAPI.swift:8-9` and `RateLimitWindow.swift:16` both cite `docs/spikes/2026-07-30-codex-usage-probe.md § "Live verification"` explicitly. The PR deliberately keeps that section binding as the fallback wire contract, so those citations remain correct and are **not** invalidated by the supersession.
- **Convention followed.** `docs/spikes/README.md` requires that a superseding finding be "marked as authoritative in place rather than deleting the original." The original static analysis and live verification are both intact; the new section is additive.
- **Cross-references resolve.** #102, #103, #104 all exist and are OPEN. Correct — this PR documents the problem, it does not implement the fix.
- **No auto-close keyword targeting them**, verified by regex against the PR body. Important here: a stray `Closes #102` would have closed an issue whose implementation hasn't started.
- **Anchor link resolves.** The new in-document link is checked against GitHub's slug algorithm (the em-dash collapses to a double hyphen — easy to get wrong by hand).
- **PII clean.** Public repo, and the diff quotes real API payloads. Scanned for emails, hostnames, account ids, UUIDs, JWTs, `rt.` refresh tokens, and absolute home paths — all sample values are redacted to `<n>` / `<epoch>` / `<plan>`.
- **CI** — `build-linux` and `build-macos` both pass on `d96977e`. Merge state `CLEAN`.

## Fixed during evaluation

- `docs/spikes/2026-07-30-codex-usage-probe.md` — the 2026-07-30 section was headed `— AUTHORITATIVE` with no qualifier, so a reader skimming to that heading could take it as the final word and never reach the supersession ~90 lines below. That is precisely the failure mode the supersession exists to prevent. Narrowed the heading to `— AUTHORITATIVE for the wire contract` and added a forward pointer. Non-behavioral, docs-only (`d96977e`).

This moved the head from `e390d96`, so CI re-ran and the verdict below is stamped against the **post-fix** tree.

## Test Execution

No `## Test Plan` section in the PR description — appropriate for a docs-only change; not treated as a defect. The repo's own gate (`swift build && ClaudeMonitor selftest`) runs in CI and passes on both platforms. No `TDD:` claim to verify, since no executing code changed.

## Notes, non-blocking

- The README's rotation note is now fairly long, and it documents a workflow (`codex import` + manual re-sync) that #102–#104 will delete outright. That is the right call for now — the docs must describe what ships today — but when #104 lands, this whole subsection should be rewritten rather than further amended. Already in #104's acceptance criteria ("README and `accounts export --help` state that Codex accounts are host-local"), so no separate follow-up issue is warranted.
- Self-review disclosure: I authored this PR and evaluated it, which is the standard Loom arrangement (GitHub blocks self-approval via the review API, hence the comment + label mechanism). I applied the same scrutiny I would to another author's diff — the two findings above (the unverified inherited citation claim, and the misleading `AUTHORITATIVE` heading) came out of that pass rather than being waved through.

PR #108, opened 2026-08-15
chore: configure Dependabot for github-actions

rjwalters:
## Summary

Adds `.github/dependabot.yml`, which the repo did not have — so no version updates were configured at all.

Scaffolded from what the repo actually contains, not a template.

## Why only `github-actions`

| Manifest | Ecosystem | Verdict |
|---|---|---|
| `.github/workflows/build.yml` | github-actions | **scaffolded** — 2× `actions/checkout@v5` |
| `package.json` (root) | npm | **excluded** — empty Loom workspace shim: `dependencies`, `devDependencies`, `peerDependencies`, `optionalDependencies` all null, and no lockfile |
| `Package.swift` | — | **excluded** — zero package dependencies by design |

`.loom/` is an installer-owned tool root (`install-metadata.json`), but carries no manifests, so nothing was excluded on that basis. Its dependencies are `/repo:update-tools`' concern, not Dependabot's.

Honest scope: the entire third-party surface is one action, already on the current major. This is upkeep — it catches future `v5.x` patches and the eventual v6 — not a fix for anything broken today.

## Choices worth reviewing

- **Majors are grouped** for Actions. The general policy keeps package-ecosystem majors ungrouped so a breaking change gets its own reviewable PR, but Actions are low-risk and per-bump PRs are noise. The comment in the file records what to do if a risk-bearing action is ever added (`exclude-patterns`, its own PR).
- **`labels: ["tier:maintenance"]`**. Every `loom:*` workflow label carries an `Applied by:` reservation (Builder, Judge, Curator, Champion, …), so all are off-limits to a bot — applying a claim label would feed automation that acts on that claim. `tier:maintenance` is unrestricted. **No label was created.**

## Also done outside this PR (repo settings, not files)

Two repository flags were enabled — they are independent of this file and can't be set by it:

- vulnerability alerts: `404 → 204` (was disabled)
- `dependabot_security_updates`: `disabled → enabled`

## Note on Loom interaction

Dependabot PRs carry no `loom:` label, so `/loom:sweep` skips them and Champion will not auto-merge them. That is safe and probably correct, but it means nothing in the Loom pipeline watches these PRs — they will sit open until triaged by a human or `/repo:deps`. Deliberately not "fixed" by labelling bot PRs; routing them into auto-merge is a policy call, not a side effect of a hygiene command.

Dependabot fires on config merge rather than waiting for the weekly interval, so any first PR should appear within minutes of this landing.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,23 @@
+version: 2
+
+# Only `github-actions` is configured, deliberately.
+#
+# The Swift package has zero package dependencies (see Package.swift), and the
+# root package.json is an empty Loom workspace shim with no dependency blocks
+# and no lockfile — neither gives Dependabot anything to update. The repo's
+# entire third-party surface is the actions referenced by .github/workflows.
+#
+# Actions are grouped including majors: they are low-risk here and reviewing
+# each bump as its own PR is noise. If a risk-bearing action is added later
+# (one coupled to an external binary or service), give it its own PR via
+# `exclude-patterns` rather than widening this group.
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-actions:
+        patterns: ["*"]
+        update-types: ["major", "minor", "patch"]
+    labels: ["tier:maintenance"]


rjwalters:
✅ **Approved!**

Single new file, 23 lines, no existing content touched.

## CI green does not validate this file

Worth stating plainly, because it is the trap with this kind of PR: `build-linux` and `build-macos` pass, but they compile Swift — they never parse `.github/dependabot.yml`. A malformed config would sail through both and then fail silently server-side, visible only in the Dependabot tab after merge. So I validated the config directly rather than treating green CI as reassurance:

- **YAML parses**, `version: 2`.
- **No unknown keys**, checked against the documented schema at both levels — top-level `updates` entry keys, and the `groups` entry keys (`patterns` / `exclude-patterns` / `dependency-type` / `update-types` / `applies-to`). `update-types: ["major","minor","patch"]` is valid inside a group.
- **`directory: "/"` is correct and complete here** — there is no `.github/actions/` directory, so no composite actions sit outside the root that a single entry would miss.
- **The label the bot would apply exists and is unrestricted** — `tier:maintenance`, description carries no `Applied by:` reservation. Dependabot silently fails to apply a nonexistent label, so this is worth confirming rather than assuming.

## PR body claims, each verified against this tree

- `2× actions/checkout@v5` — confirmed, both in `build.yml`.
- `package.json` dependency-free — confirmed, all four blocks null.
- `Package.swift` zero package dependencies — confirmed, and this one deserved care: the file *does* contain a `dependencies:` block at line 22, but it is a **target** dependency (the internal `CSQLite` target), not a package dependency. `.package(` count is **0**. A naive grep would have contradicted the PR body here; the body is right.

## On the deliberate policy departure

Grouping majors runs against the general rule that package-ecosystem majors get their own reviewable PR. For `github-actions` I agree with the exception, and the file documents it in a comment along with the escape hatch (`exclude-patterns`) if a risk-bearing action is ever added. That comment is the part that makes this maintainable — a future reader gets the reasoning, not just the setting.

## Notes, non-blocking

- `open-pull-requests-limit` is unset, so it defaults to 5. Correct for a single grouped ecosystem; no change wanted.
- The honest value here is small and the PR says so: one action, already on the current major. This is upkeep, not a fix. I'd rather approve a small honest change than see its scope inflated to justify itself.
- Scope check: this PR contains **only** the config file. The two repository flags mentioned in the description (`vulnerability-alerts`, `dependabot_security_updates`) are repo settings that no file can set, and they are correctly described as done outside this PR rather than implied to be part of it.

## Test Execution

No `## Test Plan` section — appropriate for a config-only change, not treated as a defect. No `TDD:` claim to verify, since no executing code changed. The repo's own gate (`swift build && ClaudeMonitor selftest`) passes on both platforms, which confirms the file breaks nothing, even though it cannot confirm the file is correct — hence the direct validation above.

Self-review disclosure: I authored this PR and evaluated it, the standard Loom arrangement (GitHub blocks self-approval via the review API, hence comment + label). The `Package.swift` target-vs-package distinction above is the check that would most plausibly have been waved through on a self-review; it was run.

PR #111, opened 2026-08-15
feat: read Codex usage over `codex app-server` instead of a stored token

### LINKED ISSUE #102: Read Codex usage via `codex app-server` instead of storing OAuth tokens

rjwalters:
## Summary

Adds a third way to read OpenAI/Codex usage — `codex app-server` over JSON-RPC —
and makes `pollOpenAI` prefer it. On that path the Codex CLI owns the credential
end to end: this app reads, stores, and refreshes **nothing**, which removes the
refresh-token rotation race at its root instead of re-syncing around it.

**This PR adds a path and removes none.** `OpenAIAPIClient`, the stored
access/refresh tokens, `refreshIfNeeded`'s OpenAI branch, and `accounts export`
all still exist and behave exactly as before. Retiring them is #104; per-account
`CODEX_HOME` resolution is #103 (the new client takes a `codexHome` *parameter*
defaulting to the ambient value).

## Changes

- **`Sources/CodexAppServer.swift`** (new, 830 lines / 453 non-comment):
  `CodexAppServerClient` — binary resolution, spawn, four-message handshake,
  JSONL reader, wire model, `RateLimitSnapshot` mapping, timeout + reaping.
  Conforms to `UsageProviderClient` and documents that it deliberately ignores
  the passed `ProviderCredentials` (option (a) from the issue).
- **`Sources/OAuthPoller.swift`**: `pollOpenAI` becomes a three-tier ladder —
  app-server → bearer read from `$CODEX_HOME/auth.json` at request time (never
  written back, never refreshed) → the stored credential. A merely *unavailable*
  tier falls through silently and never marks an account unhealthy; only the last
  tier's own failure sets a status. `client(for:)` still resolves `.openai` to
  the HTTP client, because that is the credential-shaped surface.
- **`Sources/SelfTest.swift`**: six new offline cases plus an opt-in `--codex`
  live probe (see Test Plan).
- **`Sources/OpenAIAPI.swift`**: two additive changes only — a
  `flatten(_:prefix:)` overload so the new client *shares* the existing
  every-depth redactor rather than duplicating it, and `Encodable` on `JSONValue`
  so a reply's `result` member can be lifted out of its envelope.
- **`Sources/UsageProviderClient.swift`**: doc comment only.
- **README / CLAUDE.md / spike doc**: document the ladder, the version
  requirement, the binary-resolution list, and the verified transport findings.

### Verified transport findings honored

Checked against `codex-cli 0.46.0` **and** `0.147.0` on this host; all recorded
in the spike doc's new §4:

| Finding | Consequence in code |
|---|---|
| Framing is newline-delimited JSON, not `Content-Length` | `CodexLineFramer` |
| Replies omit `jsonrpc`; notifications have no `id` | `CodexRPCEnvelope` does not model `jsonrpc`; id-less lines are skipped |
| `-32600`/`-32601` = unsupported method, indistinguishable from a bogus one | `isCapabilityGap` → fall back, never `.error` on the account |
| An **unauthenticated home also returns `-32600`** (new finding) | Disambiguated by `account/read`: `account == null` → `.notLoggedIn`; a non-null account means `-32600` can only be "codex too old" |
| `requiresOpenaiAuth` is `true` even when logged in (new finding) | Not used as the login signal — the spike's implied reading was wrong |
| `account/read` carries **no** account id (new finding) | `accountKey` is empty; the caller keeps its stored id, so a good reading is never discarded over identity |
| `windowDurationMins` is **minutes** | `× 60` in exactly one place, asserted directly (`10080 → 604800`) |
| Closing stdin makes the child exit 0 | Leads the shutdown sequence, ahead of SIGTERM → SIGKILL |

### Two bugs the stub-binary test caught (both would have shipped)

1. **Out-of-order replies were dropped.** The reply stash was a local of
   `awaitReply`, so a reply read while waiting on a *different* id died with that
   call. Real codex answers in order, so this would only have surfaced as a
   mysterious intermittent timeout in the field. The stash now lives in the
   session.
2. **`Process.waitUntilExit()` hangs off the main thread.** It spins a
   `CFRunLoop` on the calling thread; called from a concurrency-pool thread it
   blocked in `mach_msg` **indefinitely, after the child had already exited**
   (confirmed by stack-sampling a hung run). Reaping now polls `isRunning`.

A third hazard was found while stabilizing: ingesting pipe chunks into an
`actor` via `Task { await … }` from the readability callback **reorders them**
(unstructured tasks have no ordering guarantee), corrupting the JSONL stream.
`CodexLineStream` is therefore a lock-protected class, framing synchronously on
the callback's own serial delivery, with the invariant named at the type.

### On the "CheckedContinuation cannot be resumed twice" criterion

Satisfied **by construction: there is no continuation.** Reply-arrival, timeout,
and early-child-exit are three independent completion paths and a double resume
is a crash, so `awaitReply` polls a buffer against a deadline instead. The
exchange is sub-second, so the 15 ms wakeups are negligible, and `Task`
cancellation drops out for free.

- [x] `CodexAppServerClient` in `Sources/CodexAppServer.swift`, conforms to `UsageProviderClient`, returns a `ProviderUsageSnapshot` from `account/rateLimits/read` — verified live (`selftest --codex`).
- [x] Reader is newline-delimited, matches by `id`, tolerates a missing `jsonrpc`, ignores id-less notifications — `testCodexAppServerFraming`, `testCodexAppServerEnvelopeDecoding`, and the out-of-order stub transcript.
- [x] Kind derived from `windowDurationMins × 60` via `RateLimitSnapshot(windows:)`; `null` secondary leaves `session` nil → stored NULL — asserted on both 10080 min → weekly/604800 s and 300 min → session/18000 s.
- [x] `-32600` **and** `-32601` classify as "unsupported → fall back", not an account failure; no version parsed out of `initialize`'s `userAgent`.
- [x] Binary resolved to an absolute path from an explicit candidate list (never `/usr/bin/env`), with `CLAUDE_MONITOR_CODEX_BIN` — `testCodexBinaryResolution`, including "an override that does not resolve fails rather than silently using PATH".
- [x] Binary absent / method unsupported / home not logged in / RPC timeout each map to a distinct status; none crashes, none writes a fabricated 0% (a reply with no windows throws).
- [x] Subprocess always reaped via `defer` (stdin close → SIGTERM → SIGKILL) on success, throw, timeout and cancellation paths.
- [x] No orphaned `codex` survives repeated polls — `pgrep -f app-server` = 0 before and after 5 live runs; the stub's `EXIT` trap proves the child died in both the success and timeout cases.
- [x] Continuation double-resume impossible — no continuation exists (above).
- [x] Fallback reads the bearer via `CodexAuth.load()` at request time and never writes it back or refreshes it.
- [x] Stored-credential path, `OpenAIAPIClient`, and `accounts export` unchanged — the only edits to `OpenAIAPI.swift` are the additive `flatten(_:prefix:)` overload and `Encodable` on `JSONValue`.
- [x] `ranking.json` stays `schema: 1`; no top-level field added (`testRankingExportCarriesProvider` still passes).
- [x] `selftest` maps a captured redacted fixture offline — no network, no credentials, no `codex` — and asserts the fixture's email never reaches `rawFields`.
- [ ] **Linux `swift build && selftest`: not verified locally** — the Docker daemon is not running on this host. macOS is green; CI is authoritative. The new file imports only `Foundation` + `Glibc`/`Darwin` (added explicitly for `kill`/`SIGKILL`/`usleep`, which corelibs-foundation does not re-export) — no AppKit/SwiftUI/Combine/os.Logger.
- [x] No token, email, or raw response body logged — `grep -icE 'sk-|eyJ|@|Bearer'` over the debug.log written by 5 live polls returned 0.

### Beyond the stated criteria: multi-account misattribution guard

A single ambient `CODEX_HOME` speaks for exactly one login. On a host with two
OpenAI account rows, tiers 1 and 2 would have written the *same* login's usage to
both — a silent corruption, since the numbers look entirely plausible. Tiers 1
and 2 are therefore skipped when the reported identity **demonstrably
contradicts** the account row. The rule is deliberately asymmetric: absent
identity on either side proves nothing, so the ordinary single-account host is
unaffected. Resolving a different home per account remains #103.

## Test Plan

TDD: not test-first. This is a new transport whose wire contract had to be
established empirically first — the fixtures in `SelfTest.swift` are a redacted
capture of a real `codex-cli 0.147.0` session taken before the client was
written, and the mapping/framing/classification tests were written against that
capture alongside the implementation. The stub-binary spawn test then failed
against the first working implementation and caught the two shipping bugs listed
above, which is the substantive value the discipline is after.

- **Automated, offline (`swift build && .build/debug/ClaudeMonitor selftest`)**:
  241 checks pass, 5/5 consecutive runs, ~1.35 s. New cases cover the framing
  (split mid-line, several objects per chunk, blank lines, overflow guard), the
  envelope quirks, `-32600`/`-32601` classification and status mapping, the
  mapping (weekly 10080 min, session 300 min, null secondary, `spendControlReached`,
  a window at 100 %, a millisecond `resetsAt`, sub-limit dedupe, no-windows and
  undecodable payloads), redaction, binary resolution, and the identity guard.
- **Automated, stub binary**: one stub emits a canned transcript **out of order**
  and holds stdin open (spawn → handshake → reap, with an `EXIT` trap proving the
  child died); a second never replies (timeout → escalation → still reaped).
  Neither needs `codex` installed.
- **Manual, live (`selftest --codex`, codex-cli 0.147.0)**:
  `session —, weekly 30% (allowed), plan pro, email present: true, weekly window: 604800s, sub-limits: 1, archived fields: 45`.
- **Manual, no orphans**: `pgrep -f 'codex.*app-server'` = 0 before and after 5
  live runs.
- **Manual, leak audit**: `grep -icE 'sk-|eyJ|@|Bearer'` over the freshly written
  debug.log tail → 0.
- **Manual, not-logged-in home**: an empty `CODEX_HOME` returns
  `{"account":null,…}` then `-32600`; classified `.notLoggedIn` (`.missing`), not
  "codex too old".
- **Linux**: not run locally (no Docker daemon). CI covers it.

Closes #102


rjwalters:
CI settled green after the PR was opened, which closes the one acceptance criterion I could not verify locally (no Docker daemon on the build host):

- `build-linux` **pass** (49s) — `swift build && ClaudeMonitor selftest` on Linux, so the explicit `Glibc`/`Darwin` import for `kill`/`SIGKILL`/`usleep` and the `NSLock`/`readabilityHandler`/`Process` usage are all confirmed portable.
- `build-macos` **pass** (34s)

Every acceptance criterion in the PR body is now verified.

rjwalters:
Standing down: re-checked labels immediately before writing my verdict and found `loom:pr` already present (and my `loom:reviewing` claim removed) — another Judge's verdict raced mine. Discarding my review; not writing any label and not stamping a verdict marker.

For the record, my independent pass reached the **same** conclusion, so there is no disagreement to raise. Evidence gathered in a detached worktree at `cc3dd06`, on a host with `codex-cli 0.147.0`:

- `swift build` clean (only the two pre-existing `AnthropicAPI.swift` warnings; none in new code); `selftest` 265/265, 5/5 consecutive runs, ~1.4 s.
- `selftest --codex` live x3: `weekly 30% (allowed), plan pro, weekly window: 604800s` — the `windowDurationMins x 60` conversion confirmed against the real CLI, not just the fixture.
- `pgrep -f app-server` = 0 after 3 live runs + 5 offline runs (each spawning 2 stubs). Reaping is stdin-close -> SIGTERM -> SIGKILL in a `defer`, bounded at `gracefulExit + terminateGrace + 2` = 5 s.
- `-32600` disambiguation verified in both directions live: an empty `CODEX_HOME` yields `.notLoggedIn` (`.missing`), not "codex too old"; a bogus `CLAUDE_MONITOR_CODEX_BIN` yields `binaryNotFound`. Neither can surface as "account unhealthy" while tiers 2/3 are viable.
- Double-resume: genuinely impossible by construction — `awaitReply` has no shared completion state, and it checks the buffer *before* the deadline, so a reply already framed at deadline time wins the race. Not a busy-spin (`Task.sleep(15ms)` + `Task.checkCancellation()` at the loop head).
- Scope holds: `OpenAIAPI.swift` changes are purely additive (`flatten(_:prefix:)`, `Encodable` on `JSONValue`); `refreshIfNeeded`, `OpenAIAPIClient`, and `AccountSync` are untouched. The `#102 -> #103 -> #104` sequencing is intact.
- Portable core clean (`Foundation` + `Glibc`/`Darwin` only); Linux covered by the green `build-linux` job, which runs `swift build && selftest`.
- Leak audit: `grep -icE 'sk-|eyJ|@|Bearer'` over the freshly written `debug.log` tail = 0.

Non-blocking follow-ups for whoever touches this next (nits, not defects):

1. `noteCodexFallback` dedupes only `isCapabilityGap` failures. A host with `codex` installed but not logged in takes the `.notLoggedIn` branch, which is *not* a gap, so it writes a `[WARN]` line on every poll indefinitely. Deduping by `(credentialID, case)` would close it.
2. `decoded?.account == nil` conflates "`account/read` returned `account: null`" with "the reply did not decode as `AccountRead` at all", so a future shape change reports as `.notLoggedIn` rather than a protocol failure. Cosmetic today.
3. `reap` blocks its cooperative-pool thread (`usleep` loop) for up to 5 s in the wedged-child case. Never observed in practice; a `Task.sleep`-based variant would be strictly better.
4. The identity guard's deliberate asymmetry leaves one narrow hole: a multi-account host where an OpenAI row has a NULL `email` can still receive tier-1 numbers from the ambient `CODEX_HOME`'s login. Tier 2's account-id guard is unconditional and OpenAI rows normally carry an email, so this is narrow — #103 closes it properly.
5. `flog.info("codex app-server usage - ...")` fires from the pure `snapshot()` mapper, so offline `selftest` runs write ~14 lines into the user's real `~/.claude-monitor/debug.log`.

